PR #34, opened 2022-02-23
feat: menu component

ruslan-mart:


src/components/pickers/Menu/ActionButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+      ) {
+        const target = event.target as HTMLElement;
+
+        if (!isInsideContainer(triggerContext.menuRef.current!, target)) {
+          triggerContext.hideMenu();
+        }
+      }
+    };
+
+    document.addEventListener('mousedown', handler, true);
```
MrFlashAccount, line 69: I suppose this won't work on mobile devices, also why do you traverse over the DOM to figure out if click is outside? There are many cases when we have to use Portal API and that method won't work in that case(because menu is no more a child of MenuTrigger

PR #38, opened 2022-03-04
[CUK-31] refactor api to make more react-ish

MrFlashAccount:


src/components/actions/Button/Button.tsx
```diff
@@ -19,7 +19,7 @@ export interface CubeButtonProps extends CubeActionProps {
     | 'outline'
     | 'neutral'
     | string;
-  size?: 'small' | 'default' | 'large' | string;
+  size?: 'small' | 'default' | 'large' | (string & {});
```
JiLiZART, line 22: looks like a dirty hack

MrFlashAccount: Indeed, this hack enables autocomplete for the keywords in typescript. See more in: https://github.com/microsoft/TypeScript/issues/29729

src/components/overlays/AlertDialog/AlertDialog.stories.tsx
```diff
@@ -82,3 +89,28 @@ const Template = (props) => {
...
 Default.args = {};
+
+const ApiTemplate = (args) => {
+  const api = useAlertDialogApi();
+
+  return (
+    <Button
+      {...args}
+      onPress={() => {
+        const { promise } = api.open({
```
tenphi, line 100: Doesn't look like a convenient way to use API. It's better to use primitive properties like `onConfirm` and `onDismiss` than that.
The better way would be to return the promise directly from the `open` method. Anyway, I'm not sure we need the `close` method here. Do you know any case?

src/components/overlays/AlertDialog/AlertDialogZone.tsx
```diff
@@ -0,0 +1,67 @@
...
+export interface DialogZoneProps {
+  dialogs: Dialog[];
+}
+
+export function AlertDialogZone(props: DialogZoneProps): JSX.Element {
+  const { dialogs } = props;
+
+  return (
+    <>
+      {dialogs.map(({ props, meta }) => {
```
tenphi, line 16: Actually, it's a bad practice to use multiple modal dialogs (and AlertDialogs are always modal as far as I know). At least we can try to warn developers about the complication.

src/components/overlays/AlertDialog/AlertDialog.stories.tsx
```diff
@@ -82,3 +89,28 @@ const Template = (props) => {
...
+          children: (
```
tenphi, line 101: It's probably better to use the `content` name here to be more consistent with other apis.

tenphi (changes requested):

src/components/overlays/AlertDialog/AlertDialog.tsx
```diff
@@ -70,7 +72,7 @@ function AlertDialog(props: CubeAlertDialogProps, ref) {
           <Title>{title}</Title>
         </Header>
       ) : null}
-      {children ? <Content>{children}</Content> : null}
+      <Content>{content}</Content>
```
tenphi, line 75: Why so? It's possible that there is no content, just a header.

tenphi (changes requested):

src/components/overlays/AlertDialog/AlertDialog.tsx
MrFlashAccount: In the current version, content is a required field and I see it reasonable since if we won't have this or title required, it makes it possible to render `<AlertDialog />` without any description which is a bad practice.

tenphi (approved):

PR #68, opened 2022-04-04
fix: CUK-54 input-area doesn't stretch to the container's width

MrFlashAccount:


tenphi:
Doesn't make any sense.
It already has 100% width 'because it's a grid item. It can vary in RadioGroup 'cause it overrides `placeContent` styles which make InputArea aligned to the start.

tenphi (approved):

PR #97, opened 2022-05-04
fix(menu): make header sticky when menu is scrollable

JiLiZART:
When we set height and overflow auto to Menu component, we need show menu header.

image


tenphi:
@JiLiZART If the header is sticky the scroll should be below it.

tenphi (changes requested):
The scroll should go below the header.

tenphi (approved):

PR #110, opened 2022-06-02
chore: base component removal

tenphi:


MrFlashAccount:
The bundle size increased over the limit. You need to find out why

MrFlashAccount:
I think we should remove the deprecation of the `Base` component

src/utils/warnings.ts
```diff
@@ -45,26 +50,29 @@ export function deprecationWarning(
...
+  );
 
   // we can make deprecations even better if we add the md syntax in the console.
   // anyway, everything down below will be stripped in the production build
   console.group(`⚠️ ${PREFIX}: Deprecation in ${name}`);
   warn(
     `"${property}" is deprecated, consider better alternative: ${
       typeof betterAlternative === 'function'
-        ? betterAlternative()
+        ? (betterAlternative as () => string)()
```
MrFlashAccount, line 67: Why did you add direct casting to function?

MrFlashAccount (changes requested):

src/utils/warnings.ts
MrFlashAccount: I mean, what was the reason for casting the type? It looks like Typescript can manage it

MrFlashAccount (approved):

PR #112, opened 2022-06-17
chore(Form): stabilize behavior

tenphi:
### Form
- Fixes `rules` update
- Fixes `onSubmit`, `onSubmitFailed`, `onValuesChange` update
- Clear error on external `setFieldValue` with different value

### Other
- TooltipProvider: change `useEffect` behavior to avoid loop (I don't know exactly why)
- Update ReactAria/Select to avoid loop as well


src/components/forms/Form/Field.tsx
```diff
@@ -187,27 +187,22 @@ export function Field(allProps: CubeFieldProps) {
 
   let field = form?.getFieldInstance(fieldName);
 
+  if (field) {
```
nikolaykost, line 190: Its really not working inside useEffect?

src/components/forms/Form/useForm.tsx
```diff
@@ -141,15 +141,12 @@ export class CubeFormInstance {
     }, {});
   }
 
-  setFieldValue(
-    name: string,
-    value: any,
-    touched = false,
-    skipRender?: boolean,
-  ) {
+  setFieldValue(name: string, value: any, touched = false, skipRender = false) {
```
nikolaykost, line 144: `any` -> `unknown` or generic?

nikolaykost (changes requested):

src/components/forms/Form/useForm.tsx
tenphi: It should be `any`.

nikolaykost (approved):

PR #162, opened 2022-07-29
fix(preset.style): font-family fallback

tenphi:
### Describe changes

Fix font family fallback issues.

##### Checklist

- [ ] Pipeline is passed


.changeset/poor-rules-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cube-dev/ui-kit": patch
+---
+
+fix(preset.style): font-family fallback
```
MrFlashAccount, line 5: Please, don't write changelog as you write your commits, prefer a human-oriented style instead.

src/tasty/styles/preset.ts
```diff
@@ -10,7 +10,17 @@ function setCSSValue(
   styles[`--${styleName}`] =
     presetName === 'inherit'
       ? 'inherit'
-      : `var(--${presetName}-${styleName}, var(--default-${styleName}, inherit))`;
+      : presetName === 'default'
```
MrFlashAccount, line 13: I'm sure no one, even you after some period of time will never understand this mess of `?:${}()` symbols. Even if you wanna assign value where it's declared, I would suggest using inline self-calling arrow functions, e.g.

```typescript
styles[`--${styleName}`] = (() => {
  if (presetName === 'inherit') return 'inherit';
  // ...
})()
```

tenphi: Fixed

MrFlashAccount (approved):
LGTM 🫶

PR #163, opened 2022-08-01
Fix icon button sizes

MrFlashAccount:
##### Checklist

- [ ] Pipeline is passed

Closes:  N/A 

#

### DIFF
--- a/src/components/actions/Button/Button.tsx
+++ b/src/components/actions/Button/Button.tsx
@@ -34,16 +34,12 @@ export function provideStyles({
 
   return {
     ...DEFAULT_STYLES,
-    ...STYLES_BY_SIZE[size || 'default'],
+    ...STYLES_BY_SIZE[size ?? 'default'],
     ...(theme === 'danger' ? DANGER_STYLES_BY_TYPE : DEFAULT_STYLES_BY_TYPE)[
-      type || 'default'
+      type ?? 'default'
     ],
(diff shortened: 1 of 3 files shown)


src/components/actions/Button/Button.tsx
```diff
@@ -34,16 +34,12 @@ export function provideStyles({
 
   return {
     ...DEFAULT_STYLES,
-    ...STYLES_BY_SIZE[size || 'default'],
+    ...STYLES_BY_SIZE[size ?? 'default'],
```
tenphi, line 37: Nope, there is no false key, so the first implementation is better.

src/components/actions/Button/Button.tsx
```diff
@@ -34,16 +34,12 @@ export function provideStyles({
...
     ...(theme === 'danger' ? DANGER_STYLES_BY_TYPE : DEFAULT_STYLES_BY_TYPE)[
-      type || 'default'
+      type ?? 'default'
```
tenphi, line 39: The same is here

src/components/actions/Button/Button.tsx
MrFlashAccount: As well as any falsy values like `''` or `0`. I think using proper syntax introduced exactly for these situations is better than using a kinda hacky syntax. That's why I replaced `||` with `??`

tenphi (approved):

PR #175, opened 2022-08-12
Add suffix and prefix in Field. Reduce delay in tooltip

MrFlashAccount:
##### Checklist

- [ ] Pipeline is passed

Closes:  N/A 

#

### DIFF
--- a/src/components/forms/FieldWrapper.tsx
+++ b/src/components/forms/FieldWrapper.tsx
@@ -12,6 +12,7 @@ import { Props, Styles, tasty } from '../../tasty';
 import { TooltipProvider } from '../overlays/Tooltip/TooltipProvider';
 import { InfoCircleOutlined } from '@ant-design/icons';
 import { wrapNodeIfPlain } from '../../utils/react';
+import { Space } from '../layout/Space';
 
 const FieldElement = tasty({
   qa: 'Field',
@@ -67,9 +68,11 @@ const MessageElement = tasty({
 });
 
 export type CubeFieldWrapperProps = {
-  as: string;
+  as?: string;
   labelPosition: LabelPosition;
   label?: string;
+  labelSuffix?: ReactNode;
+  labelPrefix?: ReactNode;
   labelStyles?: Styles;
   styles?: Styles;
   /** Whether the input is required */
@@ -84,14 +87,17 @@ export type CubeFieldWrapperProps = {
   /** Styles for the message */
   messageStyles?: Styles;
   /** The description for the field. It will be placed below the label */
-  description?: string;
+  description?: ReactNode;
   Component?: JSX.Element;
   validationState?: ValidationState;
   requiredMark?: boolean;
   tooltip?: ReactNode;
+  extra?: ReactNode;
+  isHidden?: boolean;
+  necessityLabel?: ReactNode;
 };
 
-function FieldWrapper(props, ref) {
+function FieldWrapper(props: CubeFieldWrapperProps, ref) {
   const {
     as,
     labelPosition,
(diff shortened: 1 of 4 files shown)


src/components/forms/FieldWrapper.tsx
```diff
@@ -112,6 +119,8 @@ function FieldWrapper(props, ref) {
     requiredMark = true,
     tooltip,
     isHidden,
+    labelPrefix,
```
tenphi, line 122: @MrFlashAccount I wouldn't add `labelPrefix` 'cause it violates Design principles. I can't imagine any case for that.

src/components/forms/FieldWrapper.tsx
```diff
@@ -126,26 +135,41 @@ function FieldWrapper(props, ref) {
...
+            <Space placeItems="baseline" gap="0.5x">
+              <div>{labelPrefix}</div>
+              <div>{label}</div>
+              <div>{labelSuffix}</div>
+            </Space>
+
+            <Text preset="t3">{extra}</Text>
+          </Grid>
+        ) : (
+          <Space placeItems="baseline" gap="0.5x">
```
tenphi, line 154: There is no need to introduce an additional element if `suffix` is no presented.

src/components/forms/FieldWrapper.tsx
```diff
@@ -126,26 +135,41 @@ function FieldWrapper(props, ref) {
...
+            <div>{labelPrefix}</div>
+            <div>{label}</div>
+            <div>{labelSuffix}</div>
+          </Space>
+        )}
+        {tooltip ? (
```
tenphi, line 160: Looks like `tooltip` is not compatible with `extra` and `labelSuffix` props. Can you fix it? It should go right after the label text.

tenphi (changes requested):
Let's fix some issues.

src/components/forms/FieldWrapper.tsx
MrFlashAccount: Am I right that this should be right between the `label` and the `labelSuffix` ?

src/components/forms/FieldWrapper.tsx
MrFlashAccount: What design principles does it violate though?

tenphi: The labels and inputs should go with the pattern. We can't move labels and inputs inside the grid. It would complicate the navigation.

src/components/forms/FieldWrapper.tsx
MrFlashAccount: I simplified the label area in the last commit. However, I still doubt the idea of removing __unnecessary__ wrapper around the label since it adds additional logic without any functional and/or non-functional impact. Can you explain the reason behind your advice?

src/components/forms/FieldWrapper.tsx
MrFlashAccount: Thanks! fixed.

src/components/forms/FieldWrapper.tsx
tenphi: I don't like to see a mess in markup. But in the case of additional wrappers, it almost always better to reduce them so you have better control over their position.
Additional wrappers can also create unnecessary regions for readers which are not intended.

src/components/forms/FieldWrapper.tsx
MrFlashAccount: Good point, thanks!

src/components/forms/FieldWrapper.tsx
MrFlashAccount: Not sure it's applicable to our situation

tenphi (approved):

PR #177, opened 2022-08-15
fix(Field): respect onSelectionChange event of ComboBox

tenphi:
https://cubedevinc.atlassian.net/browse/CUK-123

### DIFF
--- a/src/components/forms/Form/Field.tsx
+++ b/src/components/forms/Form/Field.tsx
@@ -84,7 +84,8 @@ function getValueProps(type, value?, onChange?) {
   } else if (type === 'ComboBox') {
     return {
       inputValue: value != null ? value : '',
-      onInputChange: onChange,
+      onInputChange: (val) => onChange(val, true),
+      onSelectionChange: onChange,
     };
   } else if (type === 'Select') {
     return {
@@ -300,7 +301,7 @@ export function Field<T extends FieldTypes>(allProps: CubeFieldProps<T>) {
     validateTrigger = defaultValidateTrigger;
   }
 
-  function onChangeHandler(val) {
+  function onChangeHandler(val, dontTouch) {
     const field = form.getFieldInstance(fieldName);
 
     if (shouldUpdate) {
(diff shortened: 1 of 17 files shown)


src/components/forms/Form/Field.tsx
```diff
@@ -300,7 +301,7 @@ export function Field<T extends FieldTypes>(allProps: CubeFieldProps<T>) {
     validateTrigger = defaultValidateTrigger;
   }
 
-  function onChangeHandler(val) {
+  function onChangeHandler(val, dontTouch) {
```
MrFlashAccount, line 304: Maybe we should add types here?

tenphi: Do you think that `any` will look better here? :)
It would be too difficult to retrieve the type of field here.

MrFlashAccount (approved):

src/components/forms/Form/useForm.tsx
```diff
@@ -85,7 +86,11 @@ export class CubeFormInstance<
 
       flag = true;
 
-      field.value = newData[name];
+      if (!inputOnly) {
+        field.value = newData[name];
+      }
+
+      field.inputValue = newData[name];
```
nikolaykost, line 93: I think it brings complication in behavior of other components. Like in case we have `<Input />` inside form, then when we get `field` instance of data we got two sources of data `value` and `inputValue` it really looks confusing.

src/components/pickers/ComboBox/ComboBox.tsx
```diff
@@ -294,6 +295,7 @@ function ComboBox<T extends object>(props: CubeComboBoxProps<T>, ref) {
         {suffix}
         {!hideTrigger ? (
           <TriggerElement
+            data-qa="ComboBox_Trigger"
```
nikolaykost, line 298: ComboBoxTrigger? There is no any `underscored` data-qa attributes in whole project.

src/components/forms/Form/useForm.tsx
nikolaykost: Why we need two sources of data? What case it solves? 

Can we just switch behavior by attribute from child component?

What cases I see for `ComboBox` 
1. When it behaves like normal `Input` with ability to suggest some variants.
2. Like `Select` where entered data is restricted by set of suggested variants.

What if we just need switch on witch event we read data `onInputChange`/`inputValue` or `onSelectionChange`/`selectedKey`.
How we can detect on witch? Simply by `child.props.allowsCustomValue`

nikolaykost (approved):
I need some explanation, what cases this changes solves.
Also I have some suggestions above.

src/components/forms/Form/useForm.tsx
tenphi: That's exactly what I did. But we still need value to sync.

src/components/pickers/ComboBox/ComboBox.tsx
tenphi: Good point.

PR #206, opened 2022-09-06
CUK-128 fix(Menu): keyboard hover is not visible.

nikolaykost:
##### Checklist

- [ ] Pipeline is passed

Closes: CUK-128 

#


MrFlashAccount:
We should add a story with `play` function to describe this bug

nikolaykost:
Play functions can emulate keyboard navigation?

MrFlashAccount:
@nikolaykost ofc

tenphi (approved):

src/components/pickers/Menu/MenuItem.tsx
```diff
@@ -82,6 +83,7 @@ export function MenuItem<T>(props: MenuItemProps<T>) {
       <StyledMenuItem
         {...mergeProps(menuItemProps, hoverProps)}
         ref={ref}
+        data-qa={itemProps.qa ? `MenuItem-${itemProps.qa}` : `MenuItem-${key}`}
```
MrFlashAccount, line 86: I'm not sure it's a good idea to override a user-specified `qa` attribute

src/components/pickers/Menu/Menu.stories.tsx
```diff
@@ -253,6 +253,18 @@ export const MenuSelectableMultiple = (props) => {
   });
 };
 
+MenuSelectableMultiple.play = async ({ canvasElement }) => {
+  const { getByTestId } = within(canvasElement);
+
+  await userEvent.tab(getByTestId('Menu'));
+
+  await expect(getByTestId('MenuItem-1')).toHaveFocus();
```
MrFlashAccount, line 261: Running expect isn't necessary since we'll see these changes on the screenshot

MrFlashAccount (changes requested):

src/components/pickers/Menu/Menu.stories.tsx
nikolaykost: You mean don't run this check at all?

MrFlashAccount: Yep, but before make sure that you've captured the focused state in chromatic :)
If you wanna add a test to make sure the component is focused, add a unit test )

So the logic is:
1) If you're testing a visual state - use the storybook + play function
2) Otherwise, use unit tests

src/components/pickers/Menu/MenuItem.tsx
nikolaykost: `itemProps` is a props of `MenuButton`, also this code share qa with `li` item
When you set `qa` prop in `<Menu.Item key="foo" qa="foo-qa">....` you got 
`<li data-qa"MenuItem-foo-qa"><button data-qa="foo-qa">`

src/components/pickers/Menu/Menu.stories.tsx
MrFlashAccount: @nikolaykost you removed only one expect of 2 :)

nikolaykost: I don't understand. Why we need to write play function but don't use expect in that functions?

tenphi: @nikolaykost AFAIK it's because chromatic will snapshot states of play function. If it changes it will let us know.

MrFlashAccount (approved):

PR #210, opened 2022-09-09
[CUK-5] Field & FieldWrapper refactoring

MrFlashAccount:
##### Checklist

- [ ] Pipeline is passed

Closes:  N/A 

#


src/components/forms/TextInput/TextInput.tsx
```diff
@@ -6,21 +6,33 @@ import {
...
 ) {
-  castNullableStringValue(props);
-
+  props = castNullableStringValue(props);
   props = useProviderProps(props);
+  props = useFieldProps(props, {
+    defaultValidationTrigger: 'onBlur',
+    valuePropsMapper: ({ value, onChange }) => ({
+      onChange,
+      value: value?.toString() ?? '',
```
tenphi, line 30: Did you mean `value?.toString?.() ?? '',`?

src/components/forms/Switch/Switch.stories.tsx
```diff
@@ -42,3 +39,33 @@ export const Disabled = Template.bind({});
...
   isDisabled: true,
 };
+
+export const DisabledChecked = Template.bind({});
+DisabledChecked.args = {
+  isDisabled: true,
+  isSelected: true,
+};
+
+export const Loading = Template.bind({});
```
tenphi, line 49: It's really great that you care, but Switch doesn't have the `loading` state.

MrFlashAccount: But in real life, it does: 
CleanShot 2022-10-03 at 13 23 46@2x

src/components/forms/TextInput/TextInput.tsx
MrFlashAccount: Not sure we need to add optional chaining to `toString` since I can't remember a non `undefined` type that doesn't have `toString()` implemented

src/components/forms/Switch/Switch.stories.tsx
tenphi: Then it shouldn't

src/components/forms/TextInput/TextInput.tsx
tenphi: Bad example, but `request.body` doesn't have.
It's not so difficult to create one: `Object.create(null)`.
But yes, we can leave it as is without `?`.

PR #215, opened 2022-09-15
fix: ssr

tenphi:


src/utils/react/useViewportSize.ts
```diff
@@ -0,0 +1,53 @@
...
+interface ViewportSize {
+  width: number;
+  height: number;
+}
+
+// @ts-ignore
+let visualViewport = typeof window !== 'undefined' && window.visualViewport;
+
+export function useViewportSize(): ViewportSize {
+  let [size, setSize] = useState(() => getViewportSize());
```
nikolaykost, line 12: Seems if you dont need ‘let’ if you dont modify it

src/utils/react/useViewportSize.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  useEffect(() => {
+    // Use visualViewport api to track available height even on iOS virtual keyboard opening
+    let onResize = () => {
```
nikolaykost, line 16: Same for ‘let’

src/utils/react/useViewportSize.ts
tenphi: it's was just copied from React Aria. No need to change anything I think :)

PR #230, opened 2022-10-20
fix: transition import & SubmitError handling

tenphi:
Fix Transition import
Submit error should follow the rules:
- Don't be shown if validation is failed
- Should trigger uncaught exception if the error is not handled properly (Error object, for example)
- `SubmitError` component should show the error only if it's a valid react element or a string. For other cases, it should show "Internal error" so we don't expose unhandled errors to the user.

If Form submission is failed on the validation step, then the `onFailed` callback is not fired.


MrFlashAccount:
What was the problem? Can you explain?

MrFlashAccount:
Tests?

src/components/forms/Form/ComplexForm.stories.tsx
```diff
@@ -32,6 +32,41 @@ export default {
...
 
+const UnknownSubmitErrorTemplate: StoryFn<typeof Form> = (args) => {
+  const [form] = Form.useForm();
+
+  return (
+    <Form
+      form={form}
+      {...args}
+      onSubmit={(v) => {
+        console.log('onSubmit:', v);
```
MrFlashAccount, line 43: Please, don't use `console.log` to print out a callback's output, use `@storybook/actions` instead.

If your callback matches the default regex `on*`, the storybook will generate actions automatically. So in your case, it's enough to replace `console.log()` with `args.onSubmit(v)`:

```diff
onSubmit={(v) => {
- console.log('onSubmit:', v);
+ args.onSubmit(v);
}}
```

src/components/forms/Form/ComplexForm.stories.tsx
```diff
@@ -32,6 +32,41 @@ export default {
...
+
+        throw new Error('Unknown error');
+      }}
+      onValuesChange={(v) => {
+        console.log('onChange', v);
```
MrFlashAccount, line 51: Same here

MrFlashAccount (changes requested):

src/components/forms/Form/SubmitError.tsx
```diff
@@ -1,26 +1,33 @@
...
-  submitError?: ReactNode;
+  submitError?: unknown;
 };
 
 /**
  * An alert that shows a form error message received from the onSubmit callback.
  */
-export function SubmitError(props: CubeAlertProps) {
-  const { submitError } = useContext(FormContext) as SubmitErrorProps;
+export const SubmitError = memo(function SubmitError(props: CubeAlertProps) {
```
MrFlashAccount, line 14: It doesn't make sense to memoize this component since it accepts props from an unknown environment which makes it very easy to break the memoization

src/components/forms/Form/SubmitError.tsx
```diff
@@ -1,26 +1,33 @@
-import { ReactNode, useContext } from 'react';
+import { ReactNode, useContext, isValidElement, memo } from 'react';
 
 import { Alert, CubeAlertProps } from '../../content/Alert';
 
 import { FormContext } from './Form';
 
 type SubmitErrorProps = {
```
MrFlashAccount, line 7: I think this name isn't good bc it is named like it describes `SubmitError` props, but it is not. I consider renaming it to smth more clear.

My suggestion: 

```typescript
export type SubmitErrorProps = CubeAlertProps

type SubmitErrorContextProps = {
  submitError: unknown;
}
```

MrFlashAccount (changes requested):

MrFlashAccount:
CleanShot 2022-10-24 at 15 07 58@2x
all these cases could be easily covered with unit tests. Reach me in DM if you need some help

src/components/forms/Form/ComplexForm.stories.tsx
tenphi: It's not a matter of this PR. But I can do it separately.

MrFlashAccount: It just FYI

tenphi:
@MrFlashAccount I wasn't able to create a test that checks an uncaught rejection.

MrFlashAccount:
@tenphi feel free to reach me in DM anytime 😊

src/components/forms/Form/submit.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { renderWithForm } from '../../../test';
+import { Submit } from '../../actions';
+import { TextInput } from '../TextInput/TextInput';
+
+import { Form } from '.';
+
+describe('<Form />', () => {
+  it('should not be displayed if validation is failed on submit', async () => {
+    const onSubmit = jest.fn(() => Promise.reject('Custom Error'));
+    const onSubmitFailed = jest.fn(() => {});
```
MrFlashAccount, line 13: Just FYI, you can leave `jest.fn()` without implementation

MrFlashAccount (approved):

PR #255, opened 2022-12-01
fix(RadioGroup): gap for radio buttons

tenphi:
Reduce the distance between radio buttons inside Radio.Group

### DIFF
--- a/src/components/forms/RadioGroup/Radio.tsx
+++ b/src/components/forms/RadioGroup/Radio.tsx
@@ -35,6 +35,10 @@ const RadioWrapperElement = tasty({
     flow: 'column',
     preset: 'default',
     width: 'min-content',
+    margin: {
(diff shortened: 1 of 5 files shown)


MrFlashAccount:
Can we make these buttons smaller according to the design?

src/components/forms/RadioGroup/Radio.tsx
```diff
@@ -197,6 +201,7 @@ function Radio(props: CubeRadioProps, ref) {
         hovered: isHovered,
         button: isButton,
       }}
+      data-type={type || 'radio'}
```
MrFlashAccount, line 204: I suggest setting default values where you're destructing props instead of putting them in place of use.

MrFlashAccount (approved):
LGTM 🎉

PR #271, opened 2023-01-06
fix(Button): unify all button styles into a single object

tenphi:


src/components/forms/Form/ComplexForm.stories.tsx
```diff
@@ -6,22 +6,22 @@ import { expect } from '@storybook/jest';
 import {
   Alert,
   Block,
-  Checkbox,
-  CheckboxGroup,
-  ComboBox,
+  // Checkbox,
```
MrFlashAccount, line 9: Do we need these comments?

src/components/forms/Form/ComplexForm.stories.tsx
```diff
@@ -278,78 +270,71 @@ const Template: StoryFn<typeof Form> = (args) => {
...
-        >
-          <NumberInput label="Number field" minValue={-1} />
-        </Field>
-        <Field
-          name="slider"
-          rules={[{ required: true, message: 'This field is required' }]}
-        >
-          <RangeSlider showInput minValue={0} maxValue={100} />
-        </Field>
+        {/*<Field*/}
```
MrFlashAccount, line 273: Why did we comment out these lines?

src/tasty/tasty.tsx
```diff
@@ -184,27 +188,32 @@ function tasty<K extends StyleList, C = Record<string, unknown>>(
       propStyles = null;
     }
 
+    if (!styles || (styles && Object.keys(styles).length === 0)) {
+      styles = undefined;
+    }
+
     const useDefaultStyles = !propStyles && !styles;
+    const styleCacheKey = `${styles ? JSON.stringify(styles) : ''}${
```
MrFlashAccount, line 196: It'll run stringify on each render, are we sure it's less expensive that merging styles?

src/components/forms/Form/ComplexForm.stories.tsx
tenphi: Performance issues. In the storybook it takes too long to insert all the styles for the story.
Before that PR, we have more atomic styles for buttons. So it took less time to insert them. Not much of a difference, but it crossed the 15s limit in chromatic.
I don't want to remove this code so that I can get back to it for testing purposes. I don't see any other options.

src/tasty/tasty.tsx
tenphi: Surprisingly there are no performance issues with that but I've just improved that part to avoid stringification.

src/components/forms/Form/ComplexForm.stories.tsx
tenphi: Answered in another comment.

src/components/forms/Form/ComplexForm.stories.tsx
MrFlashAccount: Does it mean that this perf regression affects cloud perf as well?

MrFlashAccount: In my opinion, if a change causes perf regression, we should either dismiss that change or fix perf issues

tenphi: Yes, it causes, on the first load. But the following renderings will be faster because many used styles are already prepared.
The ComplexForm story was slow before. But now it has surpassed the limit.
I tested Cube Cloud. There is no measurable effect of this change. But after the release, I will have to migrate all custom buttons on Cloud to the standard ones.

MrFlashAccount (approved):

PR #276, opened 2023-01-18
feat: flattenChildren, migrate to pnpm, fix-types-WIP

shubham-kaushal:
##### Checklist

- [ ] Pipeline is passed

Closes:  N/A 

#


src/utils/react/interactions.ts
```diff
@@ -4,6 +4,7 @@ import {
   useFocus as reactAriaUseFocus,
   useFocusVisible,
 } from '@react-aria/interactions';
+import type {} from "@react-types/shared";
```
MrFlashAccount, line 7: Why here are we doing empty import?

src/utils/react/flattenChildren.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * copied from
+ * https://github.com/grrowl/react-keyed-flatten-children/
+ * as it hasn't been updated in a while and doesnt support react v17
```
MrFlashAccount, line 4: What features of react 17/18 aren't supported?

package.json
```diff
@@ -51,139 +52,140 @@
   ],
   "author": "Cube Dev Team",
   "peerDependencies": {
-    "react": ">= 17.0.0",
-    "react-dom": ">= 17.0.0",
-    "styled-components": ">= 5.3.0"
+    "react": ">= 18.2.0",
+    "react-dom": ">= 18.2.0",
+    "styled-components": ">= 5.3.6"
```
MrFlashAccount, line 57: Peer deps should be left as is since they describe the lowest boundary. That's why specifying the patch version isn't necessary and we currently not using any concurrent features in react 18

shubham-kaushal: My intention was to support React Server Component which will also require upgrading Styled-Components to v6 (the latest is v6.0.0-beta.9) which is in beta and has some breaking changes.

src/utils/react/flattenChildren.ts
shubham-kaushal: I faced some ESM build issues during the migration, and as I noticed the lib. is not updated from the last 2 years, I imported and did the fix.

src/utils/react/interactions.ts
shubham-kaushal: This PR is WIP, I've listed the pending list in the description. I might leave something here as I pushed in b/w the work.

package.json
tenphi: These newer versions of react* are not required to run UI Kit. So there is no need to bump versions just yet.

src/utils/react/interactions.ts
MrFlashAccount: Ok! Can't wait to see its final state 👀

PR #296, opened 2023-02-15
ci: migrate from yarn to pnpm

dangreen:
##### Checklist

- [x] Pipeline is passed

Closes: [CC-2094](https://cubedevinc.atlassian.net/browse/CC-2094)

#


package.json
```diff
@@ -91,6 +91,7 @@
     "@react-stately/radio": "3.6.0",
     "@react-stately/searchfield": "3.3.4",
     "@react-stately/select": "3.3.2",
+    "@react-stately/selection": "^3.11.2",
```
MrFlashAccount, line 94: Since we're adding a new dep in package.json, we have to create a new release for the ui-kit, so please don't forget to fill the changeset

MrFlashAccount (approved):

package.json
dangreen: @MrFlashAccount It is a transitive dependency of some react-stately and react-aria packages, so it's not really a new dep. It was added to escape type issues. Should I fill changeset in this case?

MrFlashAccount: I understand. Even though it's a transitive dependency, we should reflect this change for our users

dangreen: @MrFlashAccount done

tenphi (approved):

PR #300, opened 2023-02-17
Variants API, camelCase support for mods, mod parsing fixes

tenphi:
- camelCase support for mods
- Support for nested (like `:not(:last-child)`) modifiers.
- New "Variants API" for `tasty()`.


.github/workflows/pull-request.yml
```diff
@@ -13,6 +13,8 @@ jobs:
   build:
     name: 'Build & canary release'
     runs-on: ubuntu-latest
+    env:
+      NODE_OPTIONS: --max-old-space-size=4096
```
MrFlashAccount, line 17: Don't forget to add this option into other build related workflows(release, measure-size)

src/tasty/utils/cache-wrapper.ts
```diff
@@ -0,0 +1,30 @@
...
+  let cache: { string?: ReturnType<T> } = {};
+  let count = 0;
+
+  return ((firstArg: any, secondArg?: string) => {
+    const key =
+      typeof firstArg === 'string' && secondArg == null
+        ? firstArg
+        : JSON.stringify([firstArg, secondArg]);
+
+    if (!cache[key]) {
```
MrFlashAccount, line 16: I think we should consider smth like a lru cache here, wdyt?

src/tasty/tasty.tsx
```diff
@@ -140,106 +158,163 @@ function tasty<K extends StyleList, C = Record<string, unknown>>(
...
+        }
+
+        if (!styles || (styles && Object.keys(styles).length === 0)) {
+          styles = undefined;
+        }
+
+        const propStylesCacheKey = JSON.stringify(propStyles);
+        const stylesCacheKey = useMemo(() => {
+          return JSON.stringify(styles);
+        }, [styles]);
```
MrFlashAccount, line 271: I think it doesn't make sense to memoize styles here. We always pass styles object as is, without any memoization

MrFlashAccount (changes requested):

.github/workflows/pull-request.yml
tenphi: Done

src/tasty/tasty.tsx
tenphi: Sometimes we pass it as a variable (for example, while wrapping). Sometimes we pass an empty object. It will help in that cases.

src/tasty/utils/cache-wrapper.ts
tenphi: It's the old code. It was moved here.
We can add LRU cache later if you know how to make it without affecting performance.

src/tasty/tasty.tsx
MrFlashAccount: And will add overhead in all the others.

src/tasty/utils/cache-wrapper.ts
MrFlashAccount: >if you know how to make it without affecting performance.

WDYM? `lru-cache` costs O(1) for read/insert in worst case

MrFlashAccount (approved):

MrFlashAccount (approved):

PR #303, opened 2023-02-22
chore: jest upgrade

dangreen:
##### Checklist

- [x] Pipeline is passed

Closes:  N/A 

#

### Other information

Can't upgrade to React 18 because upgrade to @types/react@^18.0.0 causes the following error:

```
Error: src/tasty/tasty.tsx(215,16): error TS2590: Expression produces a union type that is too complex to represent.
```


.lintstagedrc
```diff
@@ -1,3 +1,3 @@
 {
-    "*.(js|ts|tsx|jsx)": ["prettier --write", "eslint --cache --fix", "pnpm test -- --bail --findRelatedTests"]
+    "*.(js|ts|tsx|jsx)": ["prettier --write", "eslint --cache --fix", "pnpm test -- --bail --findRelatedTests --passWithNoTests"]
```
MrFlashAccount, line 2: By the way, as we touched this file, could you remove running tests on the commit?

MrFlashAccount (approved):

.lintstagedrc
tenphi: Well, I like running tests on a commit :) Is there an issue with that?

tenphi (approved):

.lintstagedrc
dangreen: We can move test running to prepush hook.

MrFlashAccount: @dangreen agree. 

@tenphi Running tests on commit slow down your work. If you want to reduce the feedback loop, run tests while you develop a feature using `--watch` flag.

IMHO, preventing committing "inappropriate" code is a bad practice that slows development and gives nothing in return.

What if you must commit current state of work, which might not be "perfect"? There might be plenty of reasons for that.

PR #313, opened 2023-03-13
[CUK-5] New Field

MrFlashAccount:
##### Checklist

- [x] Pipeline is passed

Closes: CUK-5

#


src/components/forms/FieldWrapper/extract-field-wrapper-props.ts
```diff
@@ -0,0 +1,66 @@
...
+export function extractFieldWrapperProps<
+  Props extends CubeFieldWrapperProps,
+  FieldWrapperKeys extends keyof CubeFieldWrapperProps = keyof CubeFieldWrapperProps,
+  ActualWrapperProps extends {
+    [K in keyof Pick<
+      Props,
+      keyof CubeFieldWrapperProps
+    >]: CubeFieldWrapperProps[K];
+  } = {
+    [K in keyof Pick<
```
tenphi, line 42: A duplicate that can be turned into a separate Generic Type.

src/components/forms/Form/ComplexForm.stories.tsx
```diff
@@ -274,7 +274,7 @@ const Template: StoryFn<typeof Form> = (args) => {
           <PasswordInput label="Password field" />
         </Field>
         <Field
-          name={['select', 'one']}
+          name="select.one"
```
tenphi, line 277: Both versions should be valid. Why do we change this story? Will it stop working? I believe we use it in Cloud that way.

MrFlashAccount: I checked the cloud - it has never been used that way. Using array literals here doesn't make sense since we immediately transform an array into a string and use strings everywhere.

So I removed this use case. Using arrays here is no longer valid

tenphi: Yes, it's a one-way transformation just to make it easier to specify. We can remove it, but it would be nice to leave this feature.

MrFlashAccount: Yep, I missed these cases, reverted the change

tenphi (approved):

PR #316, opened 2023-03-16
feat(Switch): small size

tenphi:


src/components/forms/Slider/RangeSlider.stories.tsx
MrFlashAccount, line 1: can we also add at least selected state for this component?

MrFlashAccount (approved):

PR #1275, opened 2026-08-03
feat: upgrade to Tasty v3 and ESLint plugin v1

tenphi:
**Do not merge as-is** — `@tenphi/tasty` points at a snapshot build. This exists to exercise the Tasty v3 release against a real design system before it ships, and to show what the migration actually costs.

Upstream: [tenphi/tasty#227](https://github.com/tenphi/tasty/pull/227) · [tenphi/eslint-plugin-tasty#38](https://github.com/tenphi/eslint-plugin-tasty/pull/38)

## Result

| | main | this branch |
|---|---|---|
| Tests | 1139 pass | **1139 pass** |
| Type errors | 13 | **13** (byte-identical, all pre-existing and unrelated to Tasty) |
| oxlint | clean | clean |

No style values needed changing. v3's stricter directional syntax — a group naming directions takes one value, so `padding: '2x 4x top right'` becomes `padding: '2x top, 4x right'` — **is not violated anywhere in this codebase.** I verified that by running the v3 ESLint plugin over `src/`, not by grepping.

## Migration

**`getCssTextForNode` → `getCSSTextForNode`** — test helpers and the ESLint-plugin probe. Mechanical.

**`Props` is no longer exported by Tasty.** It was never a Tasty concept, just `Record<string, any>`. Worth flagging: the UI Kit **re-exported it publicly**, so this would have been a breaking change for our consumers. It is now declared in `src/props.ts` and still exported from the package root, leaving our own API unchanged. 16 files repointed.

**`Text` declares its own `block` prop.** It was inherited from Tasty's `BaseProps`, which v3 dropped as an unconsumed prop — but the UI Kit *does* consume it, as a mod and in the `'ellipsis | block'` style branch. Declaring it locally preserves the behavior.

**`Title` declares its own `inline` prop, deprecated.** Same origin, but here the prop was genuinely dead: `TitleElement` has no `inline` mod and hardcodes `display: 'block'`, so forwarding it did nothing except risk an invalid DOM attribute. Kept on the type so the public prop surface is unchanged; no longer forwarded.

## Two things worth knowing

**The ESLint plugin stays on `^0.11.2`.** Its v3 canary cannot be published: npm trusted publishing for `@tenphi/eslint-plugin-tasty` rejects the snapshot workflow's OIDC token with a 404 on the final PUT, after the build and provenance signing succeed. Adding `environment: release` to match `release.yml` did not fix it, so the allowed-workflow list on the npm package needs updating. Once that lands, this PR needs one line changed.

The old plugin does not flag v3 code (it validates the v2 at-rule spellings, and v3's kebab-case keys fall through an unconditional accept), so lint is clean either way — it just isn't validating anything v3-specific yet.

**This PR found a regression in the v3 ESLint plugin.** An earlier version of `valid-directional-modifier` reported 20 errors here, every one a false positive: `verticalAlign: 'bottom'`, `textAlign: 'left'`, `transformOrigin: 'top center'`, `transition: 'left 0.2s, top 0.2s'`. Direction words are ordinary CSS *values* for a long tail of properties, so the rule's property gate turned out to be load-bearing. Fixed upstream, with all 20 cases added as regression tests. That is the payoff for testing on a real project rather than a fixture.

---

> [!NOTE]
> **Medium Risk**
> Minor semver API change for `Text`/`Title` props plus a major styling dependency bump and intentional bundle size increase; behavior is covered by new display tests but consumers must update usages.
> 
> **Overview**
> Upgrades **`@tenphi/tasty`** to v3 and **`@tenphi/eslint-plugin-tasty`** to v1, with the mechanical renames (`getCSSTextForNode`) and a local **`Props`** type in `src/props.ts` so the package still re-exports it after Tasty dropped it.
> 
> **Breaking API cleanup:** removes **`block`** on `Text` and **`inline`** on `Title` in favor of **`display`**; `Text` ellipsis still forces block display. Docs and **`display-prop.test.tsx`** cover the migration.
> 
> Registers **12 color tokens** in `tasty.config.ts` (were only in augment types), fixes Storybook **`inset`** to v3 comma syntax (`'2x bottom, 4x left'`), and **raises size-limit budgets** (~462 kB full / ~123 kB tree-shake) for Tasty v3’s larger bundle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e4ca4ebb204fe9909535595bd5b14d8c0f2f82d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tenphi:
## Size check fails — and it is a real finding, not noise

I measured both sides locally rather than bumping the limit, because this is the number the release should be judged on.

| Entry | main (tasty 2.11.0) | v3 canary | Δ | limit |
|---|---|---|---|---|
| All | 446.93 kB | 450.33 kB | **+3.40 kB** | 460 — passes both |
| Tree shaking (just a Button) | 117.61 kB | 121.38 kB | **+3.77 kB** | 118 — **fails by 3.38 kB** |

So Tasty v3 costs a consumer roughly **3.4–3.8 kB gzipped**, and it lands on an entry that had only 0.39 kB of headroom.

The dominant contributor is dev-warning strings. v3 adds a fair amount of diagnostic prose — the directional-syntax warning, the handler-displacement warning, the chunk-conflict warning — and those live in modules that ship in every bundle. They **cannot be stripped**, because `isDevEnv()` is deliberately evaluated at runtime with bracket notation so one build works in both dev and production. Tasty's own `pnpm size` needed its limits raised for the same reason (+1.7 to +2.1 kB across its bundles).

I have deliberately **not** raised `.size-limit.cjs` here. Three options, in the order I would consider them:

1. **Trim the warning prose upstream.** The messages are long by design — they explain *why*, not just *what*. Shortening the three longest would likely recover most of the 3.8 kB.
2. **Gate the strings behind a build-time flag** so bundlers can drop them, accepting that Tasty then needs separate dev/prod builds — a real architectural change, since the current single-build design is intentional.
3. **Raise the limit** and accept the cost.

That is a call for the Tasty release, not for this PR, which is why I left the check red rather than papering over it.

---

Everything else is green: **Tests & lint pass**, **Build & canary release passes**, CodeQL and Snyk pass, Storybook publishes. Type-error count is byte-identical to main (13, all pre-existing).

tenphi:
Limit bumped in 83a3ec78 — `Tree shaking (just a Button)` goes from 118 kB to **123 kB**, so both size entries pass:

| Entry | main | this branch | limit |
|---|---|---|---|
| All | 446.93 kB | 450.33 kB | 460 kB |
| Tree shaking (just a Button) | 117.61 kB | 121.38 kB | 123 kB |

Headroom stays deliberately small (~1.6 kB) so real bloat still trips the budget, matching the file's existing philosophy. The old comment on that entry had already called it — *"only ~370 B of headroom, so this one is likely to trip next even though it passes today."*

Also merged main in, so this is up to date with #1276.

The +3.77 kB is Tasty v3's new dev diagnostics, which ship in every bundle because `isDevEnv()` is evaluated at runtime so a single build serves dev and production. Recording it here rather than treating it as noise: if that cost is ever worth clawing back, shortening the three longest messages upstream is the cheapest lever, and moving to a build-time flag is the thorough one.

Remaining checks are the two Chromatic gates, which need a human to accept or reject the visual diffs — [build #3645](https://www.chromatic.com/build?appId=626ee6db91d1c8004a912249&number=3645), 916 stories. That is the last unverified thing about this migration: unit tests and typecheck both came back identical to main, so I would expect zero visual changes, but only Chromatic can confirm it.

tenphi:
Both packages are released, so this is off the canaries and onto real versions:

| | was | now |
|---|---|---|
| `@tenphi/tasty` | `0.0.0-snapshot.cdc961c` | **`^3.0.0`** |
| `@tenphi/eslint-plugin-tasty` | `^0.11.2` | **`^1.0.0`** |

The changeset is now a `minor` and no longer says "not for release" — this is mergeable as a real upgrade.

Verified against the released versions:

- 1139 tests pass, 1 skipped
- 13 type errors, all pre-existing on `main`
- **0** tasty lint errors; the 8 remaining warnings (`consistent-token-usage` ×7, `prefer-custom-property-syntax` ×1) all pre-date this branch
- `All` 450.81 kB / 460 kB · `Tree shaking (just a Button)` 121.79 kB / 123 kB

One thing worth noting: plugin v1 lints Storybook `args.styles` and `styles={{…}}` JSX props for the first time. Story files had been silently unchecked — no detection heuristic reached them — which is exactly how the `inset: '2x bottom 4x left'` violation this PR fixes survived. Stories are now covered, and they came back clean apart from the pre-existing warnings above.

PR #1293, opened 2026-08-08
refactor(color): split ColorInput out, and add ColorPicker + ColorSwatchGroup

tenphi:
Renames the component released in 0.156.0 and gives the freed-up name to a new one.

## Why this is the right split

`ColorPicker` as shipped is a text field: a swatch, an editable color string, a popover trigger. That is an **input**. Meanwhile `Picker` already means something specific in this kit — an `ItemButton` trigger plus an overlay, no text entry (`FilterPicker` and `Select` follow the same shape). So the name was pointing at the wrong thing.

- **`ColorInput`** — the field, unchanged apart from its name.
- **`ColorPicker`** — a swatch button that opens the same popover, built on `ItemButton` exactly like `Picker`.

`DatePicker` is the counter-example worth naming: it is an input *and* carries "Picker". But it inherits that from the wider ecosystem (Ant, React Spectrum), whereas nothing outside this repo expects a color field to be called a picker — design tools use "color picker" for precisely the swatch-button-plus-popover this PR adds.

## Breaking change

`<ColorPicker value onChange label />` still compiles but now renders a **button instead of a text field** — a silent break, not a type error. The changeset says so explicitly and shows the diff. The exposure is one release old, so this is the cheapest moment to correct it.

```diff
-<ColorPicker label="Brand" format="oklch" formatMode="derive" />
+<ColorInput  label="Brand" format="oklch" formatMode="derive" />
```

`CubeColorPickerProps` now describes the trigger; the field's props are `CubeColorInputProps`. `COLOR_FORMATS`, `COLOR_SPACES`, `ColorPickerFormat` and `ColorPickerSpace` keep their names — they describe the popover both components share, so renaming them would be churn for its own sake.

## The new ColorPicker

Swatch plus the color spelled out beside it. `children` replaces that label, and `children={null}` leaves the swatch alone for toolbars and dense tables.

```jsx
<ColorPicker label="Series color" value={color} onChange={setColor} />
<ColorPicker aria-label="Series color" value={color} onChange={setColor}>{null}</ColorPicker>
```

It has no `formatMode` — with no text to reconcile, `format` alone decides how the value is written. It validates through `getValidationTheme` / `getValidationIcon` on the trigger, which is the treatment the input-component rules prescribe for trigger-shaped fields, and it registers with a form on `onChange` rather than `onBlur`.

## What is shared, and what deliberately is not

`fields/color/` now holds the color model, the channel definitions, the panel and the swatch — the real weight, used by both.

I started to extract a shared value hook too, and then removed it. The two state machines only look alike: `ColorInput` reconciles text, a last-valid fallback and three format modes against the color; `ColorPicker` has none of that. Sharing would have meant reshaping proven, tested behavior to save about twenty lines behind an abstraction neither used cleanly.

## Verification

- 1371 tests pass, including 10 new ones for the trigger and the 78 the field already had, all still green under the rename.
- `pnpm audit-docs` clean for both (bar the repo-wide `isChecked`/`scrollMargin` noise).
- `tsc` at the pre-existing 13 errors, none in these files.
- Bundle: 467.1 kB against a 469 kB budget; tree shaking unchanged at 121.8 kB. No budget bump needed.
- Both components checked in Storybook: the trigger renders across all four types, and the popover anchors under it.

---

> [!NOTE]
> **High Risk**
> Renaming and repurposing `ColorPicker` is a silent runtime break for any app still using it as a text field; new public surface area is large but well-tested.
> 
> **Overview**
> **Breaking:** the former text-field `ColorPicker` is now **`ColorInput`** (same API, including `formatMode`). The **`ColorPicker`** name is a new swatch **`ItemButton`** trigger that opens the shared popover—no text field, optional `children={null}` for compact UI, and `swatches` / `swatchColumns` for a palette under the editor.
> 
> Shared color UI moves into **`fields/color/`** (`ColorPanel`, `ColorSwatch`, model exports). **`ColorSwatchGroup`** adds a radio-style palette (hex deduping, optional `allowCustom` picker); inside a color popover, context hides custom pickers to avoid nested popovers. **`ColorInput`** omits its own trigger when already inside that popover.
> 
> Consumers of the old field must **`import { ColorInput }`**—`<ColorPicker …>` still type-checks but renders a button, not an input.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 214abdcc4c84ba7e694f0c4a2b7617e162a0f5ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tenphi:
**Scope grew — #1294 is folded in here**, so the three color components can be reviewed and verified together.

This PR now covers:

| Component | What it is |
| --- | --- |
| `ColorInput` | The field released in 0.156.0 as `ColorPicker`, renamed. API unchanged. |
| `ColorPicker` | A swatch button opening the same popover — no text field. |
| `ColorSwatchGroup` | An inline grid of preset swatches; also embeddable in `ColorPicker` via `swatches`. |

The naming rule that ties them together: **`Picker` means trigger + overlay** here, which is why the field became `ColorInput` and why the inline grid is a `Group`, not a `ColorSwatchPicker` (React Aria's term for it).

### Since the last review

- `ColorSwatchGroup` added: data-driven `colors`, `columns`, `allowCustom`, canonical-hex dedupe.
- Swatches are 20px, 16px small, 24px large, 4px apart.
- The custom option was an outline button wrapping a bordered swatch — a box inside a box. Chrome stripped; it now sits in the row as one square and takes the same selection ring.
- Popovers scroll instead of spilling. A popover caps at `50dvh`, which the panel exceeds on a short viewport, and content was rendering outside the popover's own border.
- Five Bugbot findings fixed across the branch: `cubeInputType`, placeholder styling, missing focus ring, `aria-label` outranking a visible label, and read-only leaking into an open panel.

PR #1294, opened 2026-08-09
feat(ColorSwatchGroup): add a palette of selectable swatches

tenphi:
Adds the palette half of choosing a color. **Stacked on #1293** — review that one first; the base will retarget to `main` once it merges.

## On the name

You floated `ColorRadioPicker`; I'd argue against it, and against `ColorSwatchPicker` too.

React Aria ships this exact component as **`ColorSwatchPicker`**, so that's the industry term. But we just spent #1293 establishing that in this kit **`Picker` means trigger + overlay** — that's the whole basis for splitting `ColorInput` out. `ColorPicker` (button) and `ColorSwatchPicker` (inline grid) sitting in one barrel meaning structurally different things would undo that immediately.

`ColorSwatchGroup` matches `RadioGroup` / `CheckboxGroup` / `ButtonGroup`, which is what it actually is, and pairs with the `ColorSwatch` internal. `ColorRadioPicker` mixes both metaphors and matches neither.

## What React Aria taught me

Their docs carry a constraint I'd have missed: **swatches must be unique.** Equivalent colors in different spaces are duplicates, and duplicates make selection undefined — there's no way to say which of two `#ff0000` entries is selected.

So swatches are keyed by canonical hex. `'#ff0000'`, `'rgb(255 0 0)'` and `'hsl(0 100% 50%)'` collapse to one. The same matching decides which swatch a `value` selects, so `value` needn't use the notation the swatch was written in — there's a test for each.

## Selection styling

I went with your option (b), the ring, drawn with `shadow` rather than `outline`:

```
'0 0 0 1bw #surface, 0 0 0 2bw #primary, inset 0 0 0 1bw #dark.15'
```

The first shadow is the gap, the second the ring — a 1bw ring at 1bw distance. Using `shadow` leaves `outline` free for the focus ring, which the kit uses everywhere; putting selection on `outline` would have collided with it. Option (a) shrinks the color area exactly when you most want to see it.

## The recursion guard is enforced, not documented

`allowCustom` appends a `ColorPicker`, and `ColorPicker` can contain a `ColorSwatchGroup` — so the naive version nests popovers forever. The popover provides a context flag; the group reads it and drops the escape hatch. It can't be written by hand either, not just discouraged.

Verified in the browser: 8 swatches inside the popover, **0** custom triggers, `columns={4}` producing a 4-column grid.

## Notes

- The API is `Picker`-shaped as you asked — `colors` as data, not children — while `useRadioGroup` does the work, so it's one tab stop with arrow-key navigation.
- `colors` takes `string` or `{ color, label }`; a swatch announces its color by default, which is rarely the useful thing to hear.
- `ColorPicker` gains `swatches` / `swatchColumns`.
- The hidden input reuses the kit's `HiddenInput` — my first pass set `pointer-events: none`, which broke clicking. Caught by a test.
- 1384 tests pass (11 new). Bundle 468.02 kB against the 469 kB budget, **no bump needed**; tree shaking unchanged at 121.8 kB.

---

> [!NOTE]
> **Low Risk**
> New form field components with tests and docs; read-only handling is a small behavior fix on ColorPicker with limited blast radius.
> 
> **Overview**
> Introduces **`ColorSwatchGroup`**, a data-driven grid of selectable color swatches (radio-group behavior, `Picker`-style API). Colors dedupe by canonical hex; optional **`allowCustom`** adds a trailing **`ColorPicker`** for values outside the palette.
> 
> **`ColorPicker`** gains **`swatches`** / **`swatchColumns`**, rendering a **`ColorSwatchGroup`** under the editor via **`ColorPanel`’s** new `swatches` slot. **`ColorPopoverContext`** suppresses `allowCustom` inside a popover so nested pickers cannot recurse. **`isReadOnly`** now disables the trigger and locks sliders and palette radios when the popover is already open.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48158407a4f381d3b1997f5ad2b8654db3829933. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/color-swatch-group.md
+++ b/.changeset/color-swatch-group.md
@@ -1,0 +1,23 @@
+---
+'@cube-dev/ui-kit': minor
+---
+
+Add `ColorSwatchGroup` — a grid of color swatches, one of which can be selected. It is the palette half of choosing a color, where `ColorPicker` is the freeform half.
+
+Colors go in as data rather than as children, so the API reads like `Picker` rather than `RadioGroup`, though a radio group is what it is underneath: one tab stop, arrow keys between swatches.
+
+```jsx
+<ColorSwatchGroup
+  label="Brand color"
+  colors={['#7a4dbf', '#26fcb2', { color: '#ff0000', label: 'Danger' }]}
+  columns={4}
+  value={color}
+  onChange={setColor}
+/>
+```
+
+Swatches are keyed by their canonical hex, so the same color written two ways collapses into one entry — equivalent colors would otherwise make selection ambiguous. That same matching decides which swatch a value selects, so `value` need not use the notation the swatch was written in.
+
+`allowCustom` appends a `ColorPicker` for colors outside the set, showing the current color whenever it is not one of the swatches.
+
+`ColorPicker` gains `swatches` and `swatchColumns`, which put a palette under the editor in its popover. The group drops `allowCustom` there — the escape hatch is itself a `ColorPicker`, so offering it inside one would nest popovers without end. That is enforced through context rather than documented, so the recursion cannot be written by hand either.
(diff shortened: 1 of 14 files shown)


tenphi:
Folded into #1293 — easier to review and verify the color components together. Every commit from this branch is now on `refactor/color-input-and-picker`; nothing is lost.

PR #1314, opened 2026-08-13
feat(palette): seed the brand from a color, not just a hue

tenphi:
### Describe changes

The palette could only be seeded from a **hue**, so a consumer holding a brand hex had to reverse-engineer one — and even then the brand never actually appeared. `accent-surface` was authored as a fixed `tone: '-49'` off white, so *every* accent hue resolved to roughly the same lightness: a yellow brand came out olive.

This adds **color-valued seeds** and settles the `pastel` / `saturation` overlap.

#### `accentColor` / `baseColor`

```ts
setPaletteConfig({ accentColor: '#2F5BFF', baseColor: '#7A7269' });
```

Deliberately asymmetric:

- **`accentColor`** contributes hue, saturation and **tone**. The tone is the point — it is what makes the brand fill actually *be* the color.
- **`baseColor`** contributes **hue only**. Its tone and saturation are discarded, because the chrome's lightness ladder and its 0.10–0.20 saturation factors are the design. A base color says which way the greys lean, nothing more.

Derivation lives in a new leaf module, `src/tokens/color-seed.ts`, so `palette-config.ts` can resolve a color inside `resolveConfig` without forming a cycle with `palette.ts`. That is safe because a bare-string Glaze token defaults to `lightTone: false`, making all three readings independent of the global config that `palette.ts` installs at module scope — there is a test pinning exactly that. `colorThemeSeed` now delegates to it and no longer throws on an unparseable value.

#### Fidelity: the floor is a floor, not a target

The fill is `mode: 'static'` at the color's own absolute tone with `contrast: [3, 7]`. A brand already at 5:1 — or 17:1 — is emitted **exactly as given**; only a color that cannot clear 3:1 is moved, and only as far as the floor. Measured: `#7A4DBF`, `#EF4444`, `#111827` are byte-exact in light; `#FFD400`, `#22C55E`, `#F5F5F5` are byte-exact in dark, because the floor is solved *per scheme*.

`mode: 'fixed'` was rejected: it remaps the authored tone through the light tone window and lands ~3 tone units off even when nothing binds (`#7A4DBF` emits `#8053c6`).

High contrast keeps AAA regardless of the requested color. Fidelity is a preference; that tier is selected by `prefers-contrast: more` or an explicit `data-contrast="high"`, so anyone reading it has asked for separation over brand.

#### Two things worth a reviewer's attention

**Status themes must not inherit the tone.** `extend({ colors })` would otherwise put `#danger-accent-surface` at a yellow brand's tone 88 with a red hue — a pale-pink danger button. They get the white-anchored chain restored explicitly. `special` *does* follow the brand, being the brand-on-dark CTA.

**The brand-text HC target is 9, and that number is measured.** The shipped `11` is unreachable for a saturated hue against a chromatic base: `#FFD400` in dark high contrast pinned to pure black and measured **2.23** against its base while the rest link sat at **7.07** — a hover state *less* readable than the state it intensifies. 9 is the highest target reachable across six brands in both HC schemes, holding a ~2.0 gap. A test asserts zero unreachable-contrast warnings so this cannot regress silently.

#### `pastel` and `saturation` become two paths, not two knobs that fight

Pastel is one flat chroma ceiling, so a second saturation scale on top of it only undid the evenness it exists for. Under `pastel` the seed is pinned to `100`; **setting a `saturation` therefore turns pastel off**, since tuning a saturation is the non-pastel path by definition. So `setPaletteConfig({ saturation: 55 })` keeps resolving to 55 exactly as before — **not a breaking change**. An explicit `pastel: true` written beside a saturation wins and warns in dev, but the number is kept so the toggle round-trips.

#### Fixed in passing

Under a color seed, `accent-surface` and `accent-surface-2` previously solved to the same value in high contrast, so the hover step disappeared. Re-anchoring the ramp onto the fill as a plain tone step keeps them apart.

#### The shipped palette is unchanged

`src/tokens/__snapshots__/palette.test.ts.snap` is untouched — that is the contract, and it is why the relaxed chain is conditional on a color seed rather than applied globally. Relaxing the floors unconditionally would have moved `accent-text` light tone 38.76 → 48.63 and `accent-surface` HC tone 36.08 → 51.00.

##### Checklist

- [x] Pipeline is passed

Closes: N/A

#

### Other information

**Storybook.** `Getting Started / Theming → Theme Builder` gains a **Seeded by** switch per zone (Hue / Color, plus *Follow accent* for base), requested-vs-resolved swatches that make the pastel chroma cap visible rather than mysterious, and a color-seeded `Cobalt` preset. One rule governs the panel: *a control is disabled exactly when the config field it writes is not the field in charge* — which is why the hue slider stays on screen but disabled under a color seed, showing the hue that color derived.

The three brand control clusters are now shared between `ThemeBuilder` and `Playground`, removing a near-duplicate that existed before this PR.

**Verified in the browser**, not only in tests: applying `Cobalt` renders `#accent-surface` as exactly `#2f5bff`; switching the preview to High contrast tightens it to 8.10:1 while the hover link stays above the rest link (10.19 vs 8.14); turning Pastel on visibly separates the requested and resolved swatches.

**Also fixed:** `Slate`'s `saturation: 60` became inert under the pastel pin, so it was dropped along with the comment that explained it. And saturation labels are now rounded — a saturation read off a color is a float and printed as `99.9998610927005`.

---

> [!NOTE]
> **High Risk**
> Touches global palette generation, contrast/accessibility behavior for branded colors, and a major Glaze upgrade—consumers using `accentColor`, `contrastLevel`, or direct Glaze formatting are most exposed.
> 
> **Overview**
> **Palette seeding** adds `accentColor` / `baseColor` (via new `colorSeed()`), plus `baseSaturation`, `surfaceMode: 'tinted'`, and clearer **pastel vs saturation** paths. A color-seeded accent keeps hue/chroma/tone so the brand fill can match the literal in light; status themes stay isolated from that chroma. Pinned brands use **APCA Lc floors** (page + white label) instead of WCAG ratios, with follow-up fixes so the whole ramp tracks resolved hue, label caps hold on the emitted fill, and config signatures include color strings.
> 
> **Contrast** changes so a manual `contrastLevel` only moves the normal tier while **high contrast still composes** with `data-contrast="high"` (except at level 100). **`@tenphi/glaze` bumps to 2.0.0**; color `format*` calls drop erroneous `* 100` scaling.
> 
> **UI kit**: **`ColorSwatch`** is a documented component with sizes and host-aware sizing; color fields keep a fixed **20px** swatch. **`HueSlider`** shows degrees; **`RadioGroup`** uses `border-box` for tab layouts. **ColorInput/ColorPicker** fire `onChange` per slider step and update forms without blur. **Theme Builder / Theming docs** gain Pastel/Advanced/Color tabs, export, tinted surfaces, and preview-only light/dark + HC switches. Bundle **size limit** raised to **505 kB**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1643098ecf4625cb9db85014d5be891dced056a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tenphi:
## Now built on `from` (tenphi/glaze#89)

The recipe no longer derives-and-re-seeds; it hands the color to Glaze's new `from` and pins the PR snapshot `0.0.0-snapshot.84edd5e`.

**⚠️ Do not merge until `@tenphi/glaze` is repointed at a released version.** The dependency currently resolves to a PR snapshot.

Three things improved rather than merely getting shorter:

- **The brand no longer leaks into the rest of the palette.** `from` carries an absolute saturation, so the palette-level `saturation` is no longer raised to reach a vivid brand. That was the worst side effect of the previous approach — honoring one brand color re-chromatized the neutral chrome *and* all four status themes. `#danger-accent-surface` is now identical whatever the brand is, and the test asserts that against the untouched baseline instead of a saturation-matched reference.
- **The brand text and icon are exact too, not just the fill.** `from` bypasses the light tone window per color, so `#2F5BFF` now renders `#2f5bff` as the link and the icon where it previously landed on `#3764ff`.
- **`mode: 'static'` is gone.** Exactness is scoped to light / normal contrast, which is what `from` guarantees, so the chain returns to `mode: 'fixed'` and dark maps through its window like any other fixed color.

The status-theme restore stays, and is now *more* necessary: `extend()` copies defs, so an inherited `from` would make a danger button the brand color outright rather than a washed-out version of it.

Verified live in the Theme Builder — applying `Cobalt` renders `#accent-surface`, `#accent-text-soft` and `#accent-icon` all as exactly `#2f5bff`, with the warm-grey border at hue 70.7° from `baseColor` and `danger` holding its own hue. The default-palette snapshot is still untouched.

tenphi:
## Now on the released `@tenphi/glaze@1.4.0`

[tenphi/glaze#89](https://github.com/tenphi/glaze/pull/89) is merged and released as **1.4.0**, so the dependency is off the PR snapshot and the merge blocker is cleared.

Same code either way — the built bundle measures **496,215 bytes** against both `0.0.0-snapshot.94c4efb` and `1.4.0`, and the Theme Builder renders identically (`Cobalt` still resolves `#accent-surface`, `#accent-text-soft` and `#accent-icon` all to exactly `#2f5bff`).

1846 tests pass, and the size check is inside the 498 kB budget.

Remaining before merge: the two Chromatic baselines need accepting.

PR #1344, opened 2026-08-21
perf(Root,Portal,Overlay): batch stylesheet writes into one invalidation per commit

tenphi:
Adopts tasty 3.2.0's batched style injection across the kit.

## Why

Every `insertRule()` on a live stylesheet invalidates style for that sheet's scope. Kit components inject during React's render phase, so when anything else reads layout in the same pass — a tooltip positioning itself, `TextArea` autosizing, a virtualized table measuring rows — the two interleave and the browser is forced to recalculate style between every injection. A Chrome trace of a production app showed 16 forced recalcs totalling 62 ms sandwiched between `insertRule` calls, including one 475 ms task with 52 injections and 17 recalcs.

## What changed

`configure({ batchInjection: true })` in [Root.tsx](src/components/Root.tsx), plus a batch window at each portal boundary. Windows have to be per-boundary: a commit that mounts a portal never re-renders `<Root>`, so Root's window cannot cover it.

| Window | Covers |
| --- | --- |
| [`<Root>`](src/components/Root.tsx) | App mount, and any Root re-render |
| [`<Portal>`](src/components/portal/Portal.tsx) | Tooltips — `TooltipTrigger` is its only consumer among the components |
| [`<Overlay>`](src/components/overlays/Modal/Overlay.tsx) | Popovers, modals, trays — i.e. `Dialog` and `Menu` surfaces |

Deliberately **not** covered: Board's `WidgetHost` drag clone, `LayoutPanel`, and `use-context-menu`'s 0×0 anchor. The first two measure only from event handlers, which sit outside any window in `true` mode and so are written straight through; `LayoutPanel` never measures geometry at all. A window there would buy nothing but an extra effect per portal.

No API change — no new props, no new setup for consumers.

## Why this is safe

The flush happens in `useInsertionEffect`, which React runs in the mutation phase, before any `useLayoutEffect`:

```
provider renders          -> window OPEN
  children render         -> injections queued
provider insertionEffect  -> FLUSH, window CLOSED
layout effects run        -> rules are in the sheet
```

Any commit with no window in it writes straight through, exactly as before. Turning this on can make injection cheaper; it cannot make a measurement wrong. SSR is unaffected — styles are collected as text there and the provider is inert without a `document`.

## Verification

- `pnpm test` — 97 files, 2093 pass (1 pre-existing skip). Every test using `renderWithRoot` now runs inside a batch window, so the existing suite doubles as the regression test.
- `pnpm test:browser` — 11 files, 129 pass, including the measure-heavy suites this targets: `TextArea` autosize, `DialogTrigger`, `MenuTrigger`, `Tabs`, `Board`, `DataTable`.
- Chromatic UI Tests + UI Review pass — no visual diffs, which matters because `<Overlay>` wraps every popover, modal and tray.
- `pnpm lint` clean, `pnpm build` succeeds, size +0.15% overall / +0.6% tree-shaken, both within limits.

[Root.browser.test.tsx](src/components/Root.browser.test.tsx) asserts on the write queue itself (`hasPendingStyleWrites()` mid-render), so the wiring cannot silently degrade into a no-op — the flag staying on while a provider goes missing would otherwise be invisible. Each test was verified to fail with its window removed.

## Review history worth knowing

Two bugs were found and fixed here, both by testing claims rather than reading code:

1. **Cursor Bugbot caught that this PR's original central claim was wrong.** `Popover`/`Modal`/`Tray` portal through `Overlay`'s raw `createPortal`, not `<Portal>`, so `DialogTrigger` and `MenuTrigger` — the surfaces named as benefiting most — had no window at all. Fixed by adding the `<Overlay>` window. The test that now covers it drives a real `DialogTrigger`; the original `<Portal>` test passed whether or not the overlay path was covered, so it validated the mechanism while leaving the claim untested.
2. **A StrictMode window leak in tasty itself**, found while reviewing that PR: a depth counter ended each StrictMode commit above zero, silently giving the next provider-less commit `'always'` semantics and letting its layout effect measure an unstyled box. Fixed upstream before 3.2.0 shipped; [the StrictMode test here](src/components/Root.browser.test.tsx) is the integration-level proof, and it fails against the pre-fix snapshot.

---

> [!NOTE]
> **Medium Risk**
> Touches every `Root` render and all overlay/tooltip portal commits; incorrect batch-window timing could break layout measurement, though the insertion-effect flush and new browser tests target that failure mode.
> 
> **Overview**
> **Batches runtime CSS injection** so multiple tasty `insertRule()` calls in one React commit collapse into a single style invalidation, cutting forced recalculations when components inject styles during render while overlays, autosizing, or tables measure layout in the same pass.
> 
> `Root` turns on `configure({ batchInjection: true })` and wraps the app tree in **`TastyBatchProvider`**. Because portal commits do not re-render `Root`, the same provider is added at **`Portal`** (tooltips) and **`Overlay`** (dialogs, menus, modals, trays via raw `createPortal`). Queued rules flush in **`useInsertionEffect`** before layout effects run; commits outside any provider still write immediately. **No consumer API change**; SSR unchanged.
> 
> Adds **`Root.browser.test.tsx`** to assert the queue is used (`hasPendingStyleWrites`) and that `getBoundingClientRect` in `useLayoutEffect` sees applied styles—including portal/dialog opens and a StrictMode follow-up commit.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0b76d6f41e5d3e1c3d2ffc1a04156b750addca8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tenphi:
@cursor Good catch, and it invalidated this PR's central claim rather than a detail of it — fixed in 37deb99d.

You were right: `Popover`, `Modal` and `Tray` portal through `Overlay`'s own `createPortal`, so `DialogTrigger` and `MenuTrigger` — the surfaces I named as benefiting most — got no batch window at all. Checking the rest of the portal surface, `<Portal>` turns out to cover **tooltips only**: `TooltipTrigger` is its single consumer among the components. So the placement I shipped covered the one overlay family I hadn't argued for and missed the one I had.

What changed:

- `<Overlay>` now opens a window of its own, so popovers, modals and trays are covered.
- A new test drives a real `DialogTrigger` and asserts writes were queued during the commit that mounts it. It fails without the `Overlay` window — which the existing `<Portal>` test could never do, since that one passes whether or not the overlay path is covered. It validated the mechanism while leaving the claim untested.
- The comments and changeset now name which surfaces each of the three windows covers, instead of implying blanket coverage.

Deliberately still uncovered, and now documented as such: Board's `WidgetHost` drag clone, `LayoutPanel`, and `use-context-menu`'s 0×0 anchor. None of them position themselves off a same-commit measurement, so a window buys them nothing but an extra effect. The general constraint is that windows must be opened per portal boundary, because a commit that mounts a portal never re-renders `<Root>` — worth keeping in mind for future portal sites.

PR #1378, opened 2026-08-31
refactor(form): drop the dead necessityLabel prop

tenphi:
> Stacked on #1377 — based on `feat/opt-in-necessity-indicator`, because it touches the same lines in `shared/form.ts`, `FieldWrapper/types.ts`, `Field.tsx` and three docs pages. GitHub will retarget this to `main` when #1377 merges.

## It never worked

`necessityLabel` is declared on `FieldBaseProps` and `FormBaseProps`, forwarded by `Field` and `DialogForm`, destructured by `CheckboxGroup`, exposed as a Storybook control, and documented as working in three `.docs.mdx` pages. Nothing ever read it — `FieldWrapper` doesn't destructure it, and `Label` never has.

What hid the gap: `Label` used to hold a **local variable** of the same name for the computed string.

```js
// old Label.tsx — a local, not the prop
let necessityLabel = isRequired ? requiredLabel : t('form.optional', '(optional)');
```

I checked every commit that has ever touched `necessityLabel` under `src/components/form/`, back to its introduction in #777. That local is the only "render site" in the history; the prop was never destructured from `props` in any of them. It has been a no-op in every published version.

So this deletes types, forwarding and docs — **no rendering changes**, and `pnpm test` is unchanged at 2216 passing.

## Why remove rather than implement

The task could have gone either way. Implementing it is the wrong call:

- **It's one string for two mutually exclusive states.** `(required)` and `(optional)` are different messages; a single `necessityLabel` can only be one of them.
- **It's on `FormBaseProps`.** A form-level value would stamp the same literal onto required *and* optional fields alike, which is incoherent at exactly the scope the prop is most useful.
- **i18n owns those strings now** — `form.required` / `form.optional` in `src/i18n/locales/*/uikit.json`. Rewording them everywhere is a translation override; a per-field literal prop fights localization instead of complementing it.
- **The escape hatches already exist and work.** `labelSuffix` renders content right after the label text, `extra` at the far end of the label row, and `necessityIndicator` changes the marker's shape (`'icon'` / `'label'` / `null`).

Implementing it would mean designing a new prop, not fixing a broken one — and the design that survives those four points is the one already shipping.

## Migration

Delete the prop. It was doing nothing, so nothing changes visually. For custom content beside a label use `labelSuffix` or `extra`; for the marker's shape use `necessityIndicator`.

## Checks

- `pnpm test` — 103 files, 2216 passed, 1 skipped
- `pnpm exec tsc --noEmit` — 18 errors, identical to the pre-existing baseline on `main`, none in touched files
- `pnpm audit-docs --component=Form` / `Field` / `DialogForm` — 0 issues
- `pnpm audit-defaults` — no diff
- `pnpm fix` — clean (only pre-existing `consistent-token-usage` warnings in unrelated story files)

`minor` changeset included, since it removes a public prop from the types.

---

> [!NOTE]
> **Low Risk**
> Type-only breaking removal of a no-op prop with no runtime rendering impact; risk is limited to compile failures where callers still pass `necessityLabel`.
> 
> **Overview**
> **Breaking (minor):** removes the public `necessityLabel` prop from `FieldBaseProps`, `FormBaseProps`, legacy `Field`, `Form`, `DialogForm`, and related field types—plus all destructuring, forwarding to `FieldWrapper`/children, Storybook controls, and docs references.
> 
> The prop was documented and threaded through the form stack but never consumed at render time, so **UI behavior is unchanged**; only TypeScript consumers passing `necessityLabel` need to delete it. The changeset notes alternatives: `labelSuffix` / `extra` for adjacent label content, `necessityIndicator` for marker style, and i18n keys `form.required` / `form.optional` for wording.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cfa06177fd8960f6c1c9f468f5cc9c2424ff8d2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/drop-necessity-label.md
+++ b/.changeset/drop-necessity-label.md
@@ -1,0 +1,11 @@
+---
+'@cube-dev/ui-kit': minor
+---
+
+**Breaking:** removed the `necessityLabel` prop from `FieldBaseProps`, `FormBaseProps`, `Form`, `DialogForm` and the legacy `<Field>`.
+
+It never did anything. The prop was declared on the shared field types, forwarded by `Field` and `DialogForm`, exposed as a Storybook control and documented as working — but nothing ever read it. `Label` used to hold a local variable of the same name for the computed `(required)` / `(optional)` string, which is what made the gap easy to miss on inspection; the prop itself has been a no-op in every release since it was introduced. Deleting it changes no rendering, only the types.
+
+It is not worth implementing rather than removing. It is one string for two mutually exclusive states, declared on `FormBaseProps` — so a form-level value would stamp the same literal onto required and optional fields alike. Those strings are also owned by i18n now (`form.required` / `form.optional`), so the way to reword them everywhere is a translation override, not a prop.
+
+To put custom content next to a label, use `labelSuffix` (after the label text) or `extra` (at the far end of the label row). Both already work. To change the shape of the marker itself, use `necessityIndicator`.
--- a/src/components/fields/Checkbox/CheckboxGroup.tsx
+++ b/src/components/fields/Checkbox/CheckboxGroup.tsx
@@ -62,7 +62,6 @@ function CheckboxGroup(props: WithNullableValue<CubeCheckboxGroupProps>, ref) {
     isDisabled,
     isRequired,
     necessityIndicator,
-    necessityLabel,
     label,
     extra,
     isInvalid,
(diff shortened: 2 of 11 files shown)


tenphi:
Folded into #1377 — one PR, no cross-PR conflicts. Same change, same reasoning; the `necessityLabel` removal is now the last commit there.

PR #1388, opened 2026-09-02
fix: measure auto-tooltip overflow off the commit path

tenphi:
Extracted from #1376 so it can land on its own — none of it depends on the precompiled-catalog work, and the reflow fix is worth shipping now.

## The problem

`useAutoTooltip` called `checkLabelOverflow()` synchronously inside its callback ref. React invokes callback refs during `commitAttachRef`, so reading `scrollWidth`/`clientWidth` there forced a style recalc and a layout **per element, mid-commit**.

In a Cloud profile this was the single largest forced-reflow source on the page: **122 ms of `get scrollWidth` from this one call site** — more than the entire Tasty style engine cost — with 80% of the page's 178 style recalcs JS-forced rather than frame-driven. It fires broadly, because auto tooltips are on by default and the hook backs `Button`, `Item`, `TextItem`, `LayoutHeader` and `InlineInput`.

## The fix, and the three things it broke on the way

The measurement now runs in a **microtask scheduled from the callback ref**. A microtask runs after React's entire commit — every mutation, layout effect and ref attachment — so all the labels on a page read after all of them have written, collapsing into one style and layout flush instead of forcing one each. It is still in the same frame, so there is no flash. The `ResizeObserver` now only handles later resizes.

Getting there took three follow-ups, each of which is its own commit:

- **The verdict was being cleared by the remount it causes.** Turning the verdict on mounts `TooltipProvider`, which remounts the label underneath it, and consumers like `TextItem` build their callback ref inline so React detaches and re-attaches on every render anyway. Clearing on detach therefore undid the measurement that had just been made — observer reports overflow → provider mounts → label remounts → verdict resets to `false` → no tooltip. The synchronous read used to paper over this by re-deciding on every re-attach. A transient detach now leaves the verdict alone. Reproduced against the deployed Storybook (`Content/TextItem › Overflow With Tooltip` failed with "Unable to find role=tooltip" while the span measured 545 against 200) and verified fixed by driving the same story locally.
- **`ResizeObserver`'s first delivery was too late for a loaded runner.** Observer callbacks are part of the rendering steps, so a headless browser running a thousand stories at once can delay them past the point a play function hovers. Hence the microtask rather than the observer frame.
- **Disabling auto tooltips left a stale `true`.** `isAutoTooltipEnabled` requires string children, and the label unmounts when children stop being a string — the default `Button` path, since `tooltip` defaults to `true`. Flipping the flag changes the callback ref's identity, so React detaches using the *previous* callback, which still sees auto tooltips as enabled and takes the "keep the last verdict" branch. The effect now owns the disabled case rather than depending on whether a ref happens to fire.

## Story helper

`openTooltip` waited a fixed 250 ms and hovered once; its own comment explained why that is fragile. Measuring off the commit path made the guess worse, since a label now asks for a provider a commit later than it used to. The hover is retried until the tooltip is visible — waiting on the condition rather than the clock.

The first version of that retry was worse than the flake it fixed (caught in review): retrying inside `waitFor` re-ran `unhover`/`hover` every 50 ms while `TooltipTrigger`'s open delay is 250 ms, so each retry cancelled the previous timer and the pointer never stayed hovered long enough. Each attempt now hovers **once** and parks the pointer for 1500 ms — six times the delay — before deciding the hover was wasted, three attempts, last one surfacing the real assertion error.

## Tests

`use-auto-tooltip.browser.test.tsx` is new and browser-only on purpose: jsdom reports `0` for both accessors and stubs `ResizeObserver` to a no-op, so it can see neither the overflow nor the reflow. Seven cases, including one that samples a layout effect to assert **zero** measurement reads during the commit that attaches the label refs, and one that flips children from an overflowing string to a node (it fails against the commit before its fix).

`pnpm test:browser src/components/content/use-auto-tooltip.browser.test.tsx` → 7 passed; `pnpm test` → 2,344 passed; `pnpm lint` clean.

---

> [!NOTE]
> **Medium Risk**
> Touches widely used tooltip/overflow behavior across Button, Item, TextItem, and timing-sensitive Storybook interactions; changes are well-tested but affect default UI paths.
> 
> **Overview**
> **Performance and correctness fixes** for auto-on-overflow tooltips and `TooltipProvider` mounting, plus Storybook interaction helpers and browser tests.
> 
> `useAutoTooltip` no longer reads `scrollWidth`/`clientWidth` inside its callback ref during React commit. Overflow is measured in a **queued microtask** after commit (batched layout reads), with `ResizeObserver` only for later resizes. Related logic avoids clearing the overflow verdict on transient detach when `TooltipProvider` remounts the label, and clears stale overflow state when auto tooltips are disabled (e.g. non-string children).
> 
> `TooltipProvider` replaces an effect-driven `rendered` SSR guard with **`useIsSSR`**, so the trigger mounts in a single client commit instead of remounting one commit later.
> 
> Story **`openTooltip`** accepts a lookup `() => Element`, retries hover until the tooltip is visible, and errors on detached targets. Stories were updated accordingly.
> 
> New **`use-auto-tooltip.browser.test.tsx`** (layout-read spies, remount/regression cases) and **`TooltipProvider.test.tsx`** (single-commit mount).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0abaf512f906ef822b19aa5de38ea7b017fb15c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/stories/interactions.ts
```diff
@@ -102,10 +105,46 @@ export async function openContextMenu(
...
+  // out the delay several times over before deciding the hover was wasted.
+  const ATTEMPTS = 3;
+  const PER_ATTEMPT_MS = 1500;
+
+  for (let attempt = 1; attempt <= ATTEMPTS; attempt++) {
+    // The leading unhover supplies the mouse move React Aria needs to set its
+    // interaction modality — `userEvent.hover` fires `mouseEnter` before
+    // `mouseMove`, so an un-preceded first hover is ignored.
+    await userEvent.unhover(target);
+    await userEvent.hover(target);
```
tenphi: [P1] Re-resolve the trigger before every hover attempt

The retry keeps using the Element captured before openTooltip starts. On this PR, the first unhover yields while the microtask measurement sets the overflow verdict and mounts TooltipProvider, which replaces the trigger node. The following hover and both later retries are then dispatched to that disconnected original element. This reproduces in Chromatic build 4055 for ItemButton / Auto Tooltip On Overflow and both changed TextItem stories; instrumenting the deployed story shows the first unhover on a connected node, followed by every hover with target.isConnected === false. Please make the helper re-resolve the trigger after unhover and on each attempt (for example, accept a getter/locator) instead of retrying a stale Element.

.changeset/auto-tooltip-commit-reflow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cube-dev/ui-kit': patch
+---
+
+Stop `useAutoTooltip` from measuring label overflow inside its callback ref. React invokes callback refs during `commitAttachRef`, so reading `scrollWidth`/`clientWidth` there forced a style recalc and layout per element, mid-commit — every tooltip-bearing `Button`, `Item`, `TextItem`, `LayoutHeader` and `InlineInput` paying its own reflow. The `ResizeObserver` already delivers an initial callback on `observe()`, after layout but before paint, so the measurement lands in the same frame and is batched across every observed label.
```
tenphi: [P3] Describe the microtask used by the final implementation

This release note still says the initial measurement relies on the ResizeObserver delivery, but the final code deliberately schedules a microtask because that delivery can arrive too late. Since this text is published in the changelog, please update it to describe the mechanism that actually ships.

tenphi (commented):
Two actionable findings from the deployed PR build and diff review.

src/stories/interactions.ts
tenphi: Fixed in 5d4ec481 — `openTooltip` now re-resolves the target on every attempt and accepts a lookup, and throws with the reason when it is handed a detached element. The three call sites pass getters.

Worth recording that this was the shallow half of the bug. The deeper one was that mounting `TooltipProvider` cost a *second* commit of its own — its SSR guard was a `rendered` state flipped from an effect — so the trigger remounted one task after the mount was considered finished. That is what `Menu › Comprehensive Popover Synchronization` was hitting: it resolved a button that nothing re-resolves, clicked it a few awaits later, and the click dispatched into a detached tree with no pointer event reaching the document at all. Fixed in 0abaf512 with `useIsSSR`, verified against production Storybook builds (the story fails on this branch's published build, passes on main's, passes locally once the provider settles in one commit).

.changeset/auto-tooltip-commit-reflow.md
tenphi: Rewritten in 0abaf512. The note now describes the microtask queued from the callback ref — after the whole commit, before paint, so every label reads once all of them have written — and says explicitly why it is not the ResizeObserver's first delivery (observer callbacks are part of the rendering steps, so a runner that is not producing frames can hold them past the point something asks whether the tooltip is active); the observer is described as covering later resizes only.

PR #1393, opened 2026-09-03
feat(Board): export BoardDragActiveProvider

tenphi:
## Why

`Tabs` does two things while a Board widget is being dragged:

- **spring-loads** a tab when the dragged widget hovers its header, so you can drop into it;
- **keeps every panel mounted** for the drag's duration, so the tab a widget is being pulled out of — and the board that owns the in-flight gesture — cannot unmount mid-drag and strand it.

Both read `BoardDragActiveContext`, a `createContext` private to this package. A `Board` implementation living **outside** the package therefore loses both silently, and has no way to opt in: the context is not exported, and its `false` default is indistinguishable from "no drag".

Cube Cloud has just taken its own copy of `Board` into `console-ui` so it can iterate on it without waiting on a release here (cubedevinc/cubejs-enterprise), which is precisely that case.

## What

- Export `BoardDragActiveProvider` (`{ isActive, children }`) plus `CubeBoardDragActiveProviderProps`. It publishes the drag state that `Tabs` reads, so any board can feed it.
- Route this package's own `BoardProvider` through the new component instead of the raw context, so there is one code path rather than two.
- Four specs covering what the component exists for, deliberately with no `Board` in the tree: a visited panel stays mounted while a drag is reported and unmounts when it is not, and hovering a tab header spring-loads only while a drag is reported.
- `Board.docs.mdx` → *Boards inside Tabs* now says how an external board opts in.

The raw context and `useBoardDragActive` stay internal — nothing outside needs to *read* the signal, only to publish it.

## Notes

Additive: no behaviour change for any existing consumer, and `Board` still provides the signal automatically. Full suite green (102 files / 2206 tests); `tsc --noEmit` unchanged at the 18 pre-existing errors on `main`.

---

> [!NOTE]
> **Low Risk**
> Additive API and refactor of internal wiring only; existing `Board` consumers still get drag state automatically with no migration required.
> 
> **Overview**
> **Exports `BoardDragActiveProvider`** so custom `Board` implementations outside `@cube-dev/ui-kit` can publish “widget drag in flight” the same way the built-in board does. `Tabs` already reads that signal for **spring-loaded tab switching** on hover and **keeping visited tab panels mounted** during a drag; without the provider, external boards silently miss both behaviors.
> 
> The provider is a thin wrapper around the existing internal `BoardDragActiveContext` (`isActive` + `children`). **`BoardProvider` now uses it** instead of wiring the context directly, so there is one code path. Public exports include `CubeBoardDragActiveProviderProps`; **`useBoardDragActive` stays internal**.
> 
> Docs under *Boards inside Tabs* explain when to wrap with `BoardDragActiveProvider`. **Four tests** cover panel mounting and spring-loading with the provider alone (no `Board` in the tree). Minor changeset; built-in `Board` behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c50a99e0e894be96d6f3984626faa95b00f95dad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tenphi:
Consumer: cubedevinc/cubejs-enterprise#14687, which moves `Board` into `console-ui` and is pinned to this PR's canary (`0.0.0-canary-02bbb1d`) until this releases.

PR #1397, opened 2026-09-15
feat(Select,Picker,FilterPicker): add `actions` slot to the trigger

tenphi:
## What

Adds an `actions` prop to `Select`, `Picker` and `FilterPicker`. Custom actions render **inside the trigger, to the left of the built-in clear button and dropdown caret**, and pressing one runs its handler **without opening the popover**.

```jsx
const DEFAULT_KEY = 'apple';

<Picker
  isClearable
  label="Favorite Fruit"
  selectedKey={selectedKey}
  actions={
    selectedKey !== DEFAULT_KEY ? (
      <Picker.Action
        icon={<ReloadIcon />}
        aria-label="Reset to default"
        tooltip="Reset to default"
        onPress={() => setSelectedKey(DEFAULT_KEY)}
      />
    ) : null
  }
  onSelectionChange={setSelectedKey}
>
  {items}
</Picker>
```

That not-opening-the-popover part is the whole point: it is what makes a **"Reset"** action possible — restore a default selection rather than empty it, which is something the clear button cannot express.

`ItemAction` is re-exported as `.Action` on each component (`Picker.Action`, `Select.Action`, `FilterPicker.Action`), so an action inherits the trigger's size and theme without being configured. The same static is added as **`Tab.Action`** for a tab's own `actions` slot, which previously had a documented slot but no blessed component for it. (`Tabs.Action` is unrelated — it is an `ItemButton` for the tab bar's `prefix`/`suffix` slots.)

## Why the `suffix` slot

`Item`'s grid is `icon prefix label suffix rightIcon actions`. The built-in clear button and caret live in `rightIcon`; `Item`'s own `actions` area sits *after* it, so it is the wrong side. The `suffix` slot is the one that lands immediately left of the built-ins, and it already holds a horizontal run (the user's `suffix` plus the validation icon). No change to `Item` was needed.

## Bug fixed along the way

**`Picker` and `FilterPicker` opened their popover when the built-in clear button was pressed.** `DialogTrigger` opens the popover through a `PressResponder`, whose context reaches *every* `usePress` below it — the nested clear button included. `clearValue` already calls `handleOpenChange(false)` and returns `false`, so the intent was clearly the opposite; the responder just fired afterwards. `Select`, which has no `PressResponder`, was already correct.

Both the clear button and the new `actions` slot are now wrapped in `ClearPressResponder`, which is exactly what react-aria provides for this. Regression tests cover it.

## Verification

- Unit tests across all four components: action renders before the clear button in document order, handler fires, popover stays closed, trigger itself still opens it, actions are disabled with the field, and the action is keyboard-reachable (Tab → Enter/Space activates without opening).
- Full suite green (2470 passed).
- `pnpm probe:browser` screenshots across sizes (small/medium/large), `isDisabled`, `isInvalid`, `theme="special"`, `type="clear"`, no-clear-button, two actions, and both light and dark schemes.

## Snapshot budget

`Picker` and `Select` each keep a snapshot for their `WithActions` story; `FilterPicker`'s is `NO_SNAPSHOT` since it photographs the same `ItemButton` trigger as `Picker`'s — following the existing `ClearableThemes` precedent in the same files.


tenphi:
## Review round: found and fixed a real bug

Verifying this properly turned up something the original implementation missed.

**Custom actions did nothing while the popover was open.** Pressing `Reset` with the list open only closed the list — the handler never ran. Confirmed in a real browser on all three components.

**Root cause, and it was wider than this feature.** `DialogTrigger` only asked a caller's `shouldCloseOnInteractOutside` for presses landing *outside* every `[data-popover-trigger]`. A control **inside** the trigger short-circuited to the "our own trigger, so dismiss" branch, so the predicate was never consulted and `useOverlay` swallowed the press to close itself. `Select` had the same gap in its own `useOverlay` predicate.

That means **the built-in clear button was already broken on `main`**: pressing it while the list was open did nothing at all — not even clear — even though `clearValue` explicitly calls `handleOpenChange(false)`, so closing-without-clearing was plainly not the intent. Verified against `cda687d2`.

**Fix:** ask the caller's predicate before the `data-popover-trigger` lookup, not just inside the "outside any trigger" branch; the pickers mark their trailing run with `data-trigger-action` and veto it. `DialogTrigger` stays generic — the attribute is the pickers' own convention.

After: the action runs and the popover stays open; the clear button clears and closes.

## Verification

- 6 component regression specs (2 per component), **each verified to fail without the fix**, plus one on `DialogTrigger` for the predicate ordering itself — also verified to fail against the old ordering.
- Real-browser checks: action + clear while open on all three; caret and trigger still behave; two pickers on a page still switch on one click and don't strand an open list.
- Full suites green: 2477 jsdom, 165 browser.
- Screenshot diff vs the previous commit: 20 pixels, all the loading spinner's rotation frame — no visual change.

## Also checked, no change needed

- `Escape` still closes from every state (an earlier probe suggesting otherwise was a timing artefact in the probe, not a bug).
- While the popover is open, focus stays contained in the list, so actions are mouse-reachable but not keyboard-reachable until `Escape`. That is standard popover focus containment, left as-is.
- `isReadOnly` hides the clear button but does not disable custom actions. Deliberate — a read-only field can still host an action that does not change the value — and now documented, with a pointer to pass `isDisabled` for one that does.
- React logs no DOM-nesting warning for the nested control.

PR #1412, opened 2026-09-18
feat(Form): improve modern API ergonomics and correctness

tenphi:
Modern forms currently have several traps before adoption: explicit-controller Submit buttons can target the wrong form, hook callbacks keep initial props, sibling validation can commit stale results, and input bindings lose the model's types. This change makes those workflows predictable and adds typed field descriptors while preserving legacy form behavior.

### API and behavior

- Add `form.field(path, { validate, deps, dependsOn })`, including nested tuple bindings, inferred validator values, and value compatibility checks on inputs. Built-in descriptors accept nullable API models while controls keep their existing change payloads and preserve stored nulls through mount, submission, and reset. Add `Form.useValue` / `Form.useFieldState` and typed field snapshots.
- Reject misspelled literal command/descriptor paths without enumerating every model path. Widened dynamic paths and open models remain supported. Read types reflect readonly, potentially incomplete nested data while preserving date-control classes and their methods.
- Preserve input rules when a descriptor supplies no validation rules, and give `shouldUpdate` the actual nested update shape. Default adoption reports only changed values, while separately invalidating dependent validation.
- Commit fresh hook callbacks without resetting defaults. Root callbacks remain explicit overrides and fall back to the latest hook callback. Hover documentation identifies live callbacks and initial-only options.
- Use `dependsOn` for field-triggered revalidation and `deps` for external captures. Validator reads use one values snapshot and only cancel stale in-flight work. Rule comparison detects function-source changes while preserving equivalent inline closures; `rulesKey` versions functions explicitly and skips their source comparison, while declarative constraints are always compared. `deps` alone preserves automatic detection of function replacements. Descriptors keep authored functions on rules for lazy comparison and leave public `deps` unchanged.
- Route external Submit/Reset helpers to their controller, preserve native Enter/action forms, and honor `onClick` cancellation on both buttons. Modern Submit stays enabled after validation errors by default; `disableOnInvalid` opts into disabling. Expose `state.canReset` and root `submitValues="all"`. Root `onReset` can cancel controller reset; capture is reserved for internal interception.
- Discriminate submission failures by `status`, and support explicit-controller error rendering with `renderError`.

### Review follow-up

- Reproduced and fixed stale errors after swapping plain-rule or descriptor validators. Added regressions for replacement, equivalent closures, explicit revisions, conditional dependency reads, and release/cancellation.
- Separate declared revalidation from in-flight read cancellation so the last short-circuited run does not decide future revalidation. Clear read tracking on completion and invalidation.
- Keep declarative constraints reactive with a stable `rulesKey`, while skipping function-source comparison for explicitly versioned functions (including an empty revision key). Replace dependency serialization with normalized path comparisons, remove the binding's JSON path round-trip, and move normalization outside selectors. Derive reset availability in the store's existing field scan.
- Validate descriptor paths at construction, preserve literal dotted DOM names, clarify descriptor precedence and option lifetimes, align action cancellation, and update the modern Submit default. The migration checklist explicitly calls out the changed Submit behavior for QA.
- Descriptor construction does not stringify validators or insert source strings into `deps`. Regression tests verify deferred comparison, preserved dependency arrays, and validator replacement with omitted, empty, and unchanged external dependencies.
- Keep `deps` and `dependsOn` distinct: string IDs and array-valued external captures cannot be reliably distinguished from field paths. Keep typed controller arguments on value/state hooks; descendants already avoid prop threading with `Form.useControllerContext<Model>()`, now shown in the migration guide. Keep this work in the existing PR for manual review rather than opening replacement PRs.

### Documentation and API choices

- Add a scenario guide with a recommended path for ordinary forms and explicit reasons for alternatives: binding, callbacks, subscriptions, server data, validation dependencies, conditional fields, wizards, errors, and custom controls.
- Separate legacy references from modern recipes and remove duplicated modern explanations from the migration guide and entry pages. Keep advanced defaults policies distinct from ordinary loading/reset workflows; clarify component aliases and payload selection.
- Render one Markdown source in Storybook and include it in packaged docs. The docs exporter expands the shared import, and compiled consumer examples follow the same recommended API choices. This documentation round removes no runtime APIs.

### Compatibility

The modern API intentionally changes before adoption: hook callbacks stay fresh, unknown literal paths are rejected, nested reads are readonly/incomplete, and `onSubmitFailed` receives `{ status: 'invalid', errors }` or `{ status: 'failed', error }`. External captures belong in `deps`; sibling form paths belong in `dependsOn`. Function source cannot distinguish captures or bound/native functions with identical source, so list those changing inputs in `deps`. Existing legacy forms retain their APIs. The migration guide, Form/component-author docs, and a minor changeset are included.

### Validation

- Documentation follow-up: package and Storybook builds, built-consumer type checks, lint, and Form docs audit pass. Browser review confirms all seven decision tables and section links render correctly; exported Markdown matches the source guide. Runtime suites below were run before this documentation/example-only follow-up.
- Full unit suite: 2,725 passed, one existing skip. Pure-store suite: 126 passed. Compiled modern React suite: 82 passed.
- Source and built-consumer type checks; package build. Consumer checks cover `canReset` and rejection of controller `onResetCapture`.
- Lint, Form compiler diagnostics ratchet, defaults audit, and Form docs audit pass.
- Previous implementation CI passed on `99dfb9a8`: React 18/19 (including compiled consumers), browser tests, full tests/lint, previews, security checks, and all 1,063 Chromatic stories unchanged.
- CI bundle measurement: All 525,316 B gzip within the existing 526 kB limit; Button unchanged at 119,482 B.

Cloud was not changed or tested. This PR is for manual review and must remain unmerged.


tenphi (commented):
Reviewed the full diff against `62d1b5fb`, ran the pure-store suite locally (116 passed), and wrote throwaway probes to check the invalidation model on both this branch and `main`. CI is green, including `test:types:consumer`, so the `@ts-expect-error` fixtures do hold against the built declarations.

The lifecycle work here is real and the test coverage is unusually good. My concerns are concentrated in two places: the new rule-signature rule has a silent failure mode, and the validation-dependency surface has grown to four overlapping knobs.

---

## 1. Swapping a validator function silently keeps the stale error — regression vs `main`

`modern/validation.ts`:

```diff
-if (typeof value === 'function') return `function:${value.toString()}`;
+if (typeof value === 'function') return 'function';
```

Every function now hashes to the same constant, so a genuinely different validator is indistinguishable from the previous one. Verified with the same probe on both branches:

```js
const reg = store.register('a', { rules: [{ validator: () => 'strict error' }] });
await store.validate();                                  // { status: 'invalid', errors: ['strict error'] }
reg.update({ rules: [{ validator: () => undefined }] }); // swap to a passing validator
await flush();
```

| branch | result |
|---|---|
| `main` | `{ status: 'unvalidated', errors: [] }` |
| this PR | `{ status: 'invalid', errors: ['strict error'] }` — and it stays there |

With `rulesKey` or `deps` it correctly becomes `{ status: 'valid', errors: [] }`, so the recovery path works — but nothing signals that it is needed. The pattern that trips on this is ordinary:

```jsx
rules={mode === 'strict' ? [{ validator: strict }] : [{ validator: loose }]}
```

The user is left looking at an error message produced by a validator that is no longer installed. `docs/modern-form-migration.md` does say "Inline function identity does not restart validation, so change `deps` when captured inputs change" — but that sentence is about *captures*, and this is about the *rule itself* changing. There is also no test for it; `dx.test.tsx:322` covers the `deps`-change case, not the bare-swap case.

Suggestion: keep `value.toString()` in the signature. It is the only automatic way to notice a different validator, it is stable for inline closures across renders (which is the churn you were avoiding), and it costs nothing you are not already paying — see the next point.

## 2. `rulesKey` lost its cost-avoidance role, and the hot path got more expensive

```diff
-const signature = prepared.rulesKey ?? rulesSignature(prepared.rules);
+const signature = rulesSignature(prepared.rules);
```

`rulesKey` used to let a consumer skip hashing the rules graph entirely. Now `rulesSignature` walks and stringifies the whole rule set on *every* registration update — i.e. every React commit, for every field — and `update()` additionally runs `JSON.stringify` twice:

```js
JSON.stringify(registration.options.dependsOn) !== JSON.stringify(prepared.dependsOn)
```

`sameDependencies` is defined fifteen lines above and a shallow-array-of-paths variant would do the job without allocating. The modern store's README states the dirty comparison "never uses `JSON.stringify`"; this quietly introduces it on the options path.

Same pattern in `use-field/use-field-binding.ts`:

```js
const pathKey = controller ? JSON.stringify(normalizePath(props.field?.path ?? name)) : '';
// ...
createModernFieldBackend(controller, JSON.parse(pathKey))
```

A serialize/deserialize round-trip per render purely to get a stable memo key, and `JSON.parse` returns `any`, so the path loses its type on the way in. `getFieldKey()` already produces a stable string — use it as the memo key and memoize the path itself alongside it.

## 3. Read-tracked dependencies are order-dependent

`record.reads` is populated from whatever the validator actually read on its **last** run, and `dependencyChanged` uses it to decide re-runs. A validator that short-circuits records nothing:

```js
if (!value) return 'Required';
return value === ctx.getValue('password') ? undefined : 'Mismatch';
```

Confirmed by probe: with `confirm` empty, changing `password` does **not** revalidate (validator call count stays at 1). Fill `confirm` in, validate once, and now the same `password` change *does* revalidate. Identical code, two different behaviours, decided by history.

`dependsOn` fixes it, but nothing tells the consumer they need it, and the bug reproduces only from a specific starting state.

The cleaner model is already half-built here: reads are sound for **cancelling the in-flight run** (you know it read a value that has since changed), and unsound for **deciding to re-run**. Split them — use `reads` only for cancellation, require a declaration for re-running. That is deterministic, removes a class of "worked yesterday" reports, and shrinks `dependencyChanged` to a declared-paths scan.

## 4. Four knobs for one question

To answer "did my validator change?", a modern field now offers: structural `rules` hashing, `rulesKey`, `deps`, `dependsOn`, and implicit read tracking. Two of these are exact duplicates (`rulesKey: 'x'` ≡ `deps: ['x']`) and two overlap (`dependsOn` ≡ read tracking, but reliable).

`deps` and `dependsOn` are English synonyms sitting on the same props object meaning different things. That is the kind of distinction a consumer gets wrong once and then copies forever.

Proposal — one prop:

```ts
/** Validator dependencies. A string or tuple is a form path; anything else is an external capture. */
deps?: readonly unknown[];
```

- `deps: ['password']` → form path
- `deps: [organizationId, checkEmail]` → external captures
- `deps: [['profile', 'name'], orgId]` → both

That removes `dependsOn`, deprecates `rulesKey` on the modern path, and leaves read tracking as a pure cancellation optimisation with no public surface. Discriminating path-vs-value at the store boundary is a few lines.

Also: `deps` and `dependsOn` have no JSDoc in `FieldCoreProps` (`src/shared/form.ts:77-78`), unlike every neighbouring prop.

## 5. Hook options have two lifetimes and the type doesn't say which

`useFormController(options)` now commits `onSubmit` / `onSubmitFailed` / `onValuesChange` after every render, while `defaultValues`, `errorPolicy`, `onListenerError`, and `onDevelopmentError` stay frozen at creation. `errorPolicy` in particular is read live inside `validateRecord` (`options.errorPolicy ?? 'first'`) but `updateCallbacks` never refreshes it — so passing it from state looks like it should work and doesn't.

One object, two lifetimes, and `UseFormControllerOptions` is a flat alias with nothing marking the boundary. Split the type (`FormCallbacks` live / store policy frozen) or at minimum JSDoc each field. The prose in the migration guide is not where a consumer will look.

Minor: the `useLayoutEffect` in `useFormController` has no dependency array, so it mutates the store's `options` on every render of the owner. Cheap, but worth a comment saying it's deliberate.

## 6. Descriptor precedence is broader than its doc

```js
provided.field ? { ...provided, ...provided.field.options, form: ..., name: ... } : provided
```

The descriptor's options override *every* matching prop, not just `form`/`name`:

```jsx
<TextInput field={form.field('a', { validateTrigger: 'onBlur' })} validateTrigger="onChange" />
// silently uses onBlur
```

The JSDoc says "takes precedence over form/name". `docs/modern-form-migration.md` is accurate; the prop doc a consumer hovers is not. Everywhere else in React the more local JSX prop wins — either flip the spread order or state the rule where it's read.

## 7. `Form.useValue` / `Form.useFieldState` ignore the controller context

`Form.Subscribe` already supports the omit-for-context convention (`Object.hasOwn(props, 'form')`). These two require the controller positionally:

```jsx
Form.useValue(form, 'email')      // now
Form.useValue('email')            // what people will reach for inside <Form>
```

Adding the overload is small and makes the "place them in leaf components" advice in the migration guide actually pleasant to follow — leaf components are exactly where threading `form` down is annoying.

## 8. Submit and Reset disagree on the event model

- `ModernSubmitButton` intercepts DOM `onClick` and honours `event.defaultPrevented`.
- `ModernResetButton` calls `root.reset()` **then** `onPress?.(event)` — so the handler runs *after* the reset, and a React Aria `PressEvent` has no `preventDefault()` anyway. Reset cannot be cancelled from the button at all.

Everything else in this kit is `onPress`. Pick one model; if cancellation is wanted, an explicit `onBeforeSubmit` / `onBeforeReset` returning `false` is clearer than relying on which of two event systems a given button happens to use.

## 9. `onReset` / `onResetCapture` don't mean what the DOM means — and mean different things in two modes

```jsx
onReset={otherProps.action == null ? undefined : otherProps.onReset}
onResetCapture={otherProps.action == null ? handleReset : otherProps.onResetCapture}
```

with `handleReset` calling `onResetCapture?.(event)` and then `onReset?.(event)` back to back. So on a controller form, `onReset` fires during capture and `onResetCapture` is just "a second `onReset`"; on a native `action` form both revert to real DOM semantics.

The capture-phase interception itself is a legitimate fix for the React Aria listeners. But exposing `onResetCapture` publicly buys nothing and adds a second DOM-named prop that isn't the DOM prop. Drop it and document `onReset` as "runs before the controller reset; `preventDefault()` cancels it".

Worth calling out explicitly in the docs too: because `handleReset` calls `stopPropagation()` unconditionally, cancelling via `preventDefault()` now means *nothing* resets — including React Aria's own listeners for any nested non-controller input. That is the intent, but it is a sharp edge.

## 10. `disableOnInvalid` still defaults to `true`

A disabled submit button is a long-standing a11y problem — a keyboard user gets no announcement of why nothing happens, and the errors they need are often below the fold. `false` (submit, validate, show the errors) is the better default.

This PR's own premise is that modern forms have no adopters yet and the API "intentionally changes before adoption". If the default is ever going to flip, this is the PR. Adding the opt-out but keeping the old default means every future consumer has to know to pass it.

## 11. `Form.Reset`'s enabled predicate scans every field on every notification

```js
state.isDirty || state.isTouched || state.submitError != null ||
Object.values(state.fields).some(f => f.errors.length || f.status !== 'unvalidated')
```

`useFormSelector` runs its selector on every store notification, so this is O(fields) per keystroke per Reset button. More importantly the rule is subtle enough (a merely-*validated* field enables Reset) that it belongs in the store as a derived `canReset`, where it can be tested once and reused by Cloud's own footers.

## 12. Smaller things

- `useFormValue` / `useFormFieldState` call `normalizePath` / `getFieldKey` **inside** the selector, so both allocate a frozen array (and `getFieldKey` a mapped array plus a join) on every notification and every render. Hoist with `useMemo`.
- `createFormField` doesn't call `normalizePath`, so `form.field([])` or `form.field(['__proto__'])` throws later, inside the input's layout effect, instead of at the call site. The factory is the right place to fail.
- `useFieldProps` sets `name: getFieldKey(field.path)`, which escapes dots. `form.field('user.email')` — one literal key — yields `props.name === 'user\\.email'`, which then feeds id generation and the rendered input. Works, surprising in devtools.
- `record.reads` is never cleared on release, so a retained record keeps the previous validator's read set and can be invalidated by paths nothing reads any more.

## 13. Scope

Ten runtime API additions, seven new exported types, and one breaking callback signature, layered on a validation-pipeline rewrite. Since this is explicitly not merging as-is, splitting it would make each piece independently reviewable and independently revertable:

1. validation dependencies + cancellation (the store work — where the risk is)
2. typed field descriptors + read types (the type work — self-contained)
3. action routing, reset capture, submit payload selection (the DOM work)

Point 1 alone is worth its own review pass; it is currently the smallest part of the diff and the largest part of the risk.

---

**Summary.** The descriptor and read-type work is good and I'd take it largely as-is. I'd block on #1 (stale error after a validator swap) and want #3 resolved one way or the other before anyone adopts this — both are silent, state-dependent failures, which is the worst category to ship into a form library. #4 and #7-#10 are the places where the API is currently harder than the problem.

tenphi (commented):
Re-reviewed at `2d587134`. Re-ran my original probes against the new head and the pure-store suite (122 passed, up from 116). CI is fully green now, including Chromatic (`1063 tests unchanged` on build 4114).

Both things I said I'd block on are fixed, and fixed the way the store wanted rather than papered over. Verified independently rather than from the diff:

| probe | before | now |
|---|---|---|
| bare validator swap, no `deps`/`rulesKey` | `{status:'invalid', errors:['strict error']}` — permanently | `{status:'valid', errors:[]}` |
| equivalent inline closure re-registered | — | field snapshot identity preserved, no churn |
| short-circuit sibling, **undeclared** | 1 run, then 1+ runs — order dependent | 1 run, then 0 runs — never |
| short-circuit sibling, **declared** | — | 2 runs, then 1 run — always |
| `canReset` fresh → edit → reset | — | `false` → `true` → `false` |

The `dependencyChanged` / `validationReadChanged` split is exactly the right shape: reads cancel in-flight work, declarations schedule re-runs, and `revalidate` is gated on `(fieldChanged || dependency)` so a read can never schedule one. That removes the whole class of state-dependent behaviour. The parameterised `it.each([false, true])` test on the password/confirm case pins both directions, which is better than what I asked for.

Also landed: `canReset` derived in `snapshot()` inside the existing record loop (no extra pass); `samePaths`/`sameDependencies` replacing both `JSON.stringify` comparisons; `getFieldKey`/`getFieldPath` replacing the stringify/parse round-trip; `invalidate()` and `complete()` clearing `record.reads`; `normalizePath` in `createFormField` so bad paths throw at the call site; unescaped `name` for literal string paths; `settled` guards on the validator getters; lifetime JSDoc on every `FormStoreOptions` / `FormCallbacks` field; `onResetCapture` typed `never`; `disableOnInvalid` flipped to `false`; Reset moved to `onClick` with cancellation, matching Submit.

Answering #7 with `Form.useControllerContext<Profile>()` + a documented leaf-component example rather than a context overload is the better call — an overload couldn't infer `T` from context anyway.

---

Three small things, none blocking.

### `rulesKey` masks structural rule changes again

Restoring `prepared.rulesKey ?? rulesSignature(prepared.rules)` fixed the cost problem, but it also restored the masking. Verified:

```js
const reg = store.register('a', { rulesKey: 'k', rules: [{ min: 2 }] });
await store.validate();                              // valid, value 'abc'
reg.update({ rulesKey: 'k', rules: [{ min: 99 }] }); // -> still { status: 'valid', errors: [] }
```

This is pre-existing `main` behaviour, it is now documented ("bypassing rule comparison. Update when any rule changes"), and `lets rulesKey explicitly version all rules` asserts it — so it's a deliberate contract, not a slip. Folding `isRequired` into the key in `fieldRegistrationOptions` covers the most common case.

It's still reachable in an ordinary combination: `<TextInput rulesKey={orgId} rules={[{ min: minLen }, { validator: checkAgainstOrg }]} />`, where `minLen` comes from props. `orgId` is stable, so a `minLen` change is invisible.

There's an option that gets both properties — always run `rulesSignature`, but emit the constant `'function'` instead of `value.toString()` when `rulesKey` or `deps` is present:

```js
const signature = rulesSignature(prepared.rules, { hashFunctions: !prepared.rulesKey && !prepared.deps });
```

Declarative constraints are always compared (cheap — no `toString`), and the expensive source hashing is paid only when the consumer has given you no explicit versioning. That is strictly better than either of the two positions this PR has held, and it makes `rulesKey` mean "version my closures" again instead of "turn off rule comparison".

### `validate.toString()` lands in the public `deps`

```js
form.field('email', { deps: ['org-1'], validate: (value) => value?.includes('@') ? undefined : 'Invalid email' })
// field.options.deps === ['(value) => value?.includes("@") ? undefined : "Invalid email"', 'org-1']
```

The mechanism is sound (the rule adapter has fixed source, so the authored function needs its own channel), but it is recomputed on every `form.field()` call — i.e. every render of every descriptor-bound input with a validator — and `FormField.options` is public, so the source string is visible to anyone inspecting a descriptor.

Putting the authored function on the rule object instead would make it lazy and keep `deps` clean, since `rulesSignature` already stringifies any function it finds in the rule graph:

```js
rules: [...(registration.rules ?? []), { validator: adapter, authored: validate }]
```

Then it costs nothing when `rulesKey` short-circuits, and `deps` stays what the consumer wrote.

### `deps` / `dependsOn`

Still both present. The docs now need an explicit disambiguator — "strings in `deps` are values, while strings in `dependsOn` are field paths" — which is itself the evidence that the two names don't carry the distinction. Not worth reopening now that the semantics are clean and tested; noting it as the one place I'd still take the single-prop union if this surface is ever revised.

### One doc line

Legacy Submit disables on invalid, modern Submit no longer does. The actions section says so, but the per-form migration checklist doesn't — step 7's testing list is where someone migrating a form will actually look, and an always-enabled submit button is a visible change to QA.

---

The store-side model is in good shape now — the parts I was worried about were silent and state-dependent, and they aren't any more.

PR #1415, opened 2026-09-21
fix(overlays): stop a closing popup from swallowing Escape

tenphi:
## The bug

A `Select`, `Picker` or popover `DialogTrigger` inside a Dialog leaves the Dialog open on `Escape` once its own popup has been used. Reported as CUB-4839 after the 0.179.0 bump made a pre-existing race deterministic (0.177 → 2/8, 0.178 → 1/8, 0.179 → 8/8).

The bump is not the cause. It only shifted the timing of a window that has always been there.

## Impact

The reported scenario — Select in a Dialog, keyboard pick, one Escape — measured in real Chromium over 10 fresh iterations:

| | one Escape closes the Dialog |
|---|---|
| `main` | **0 / 10** |
| this PR | **10 / 10** |

## Three swallows, not one

Each of these is on its own enough to keep the Dialog open, which is why the symptom survived partial fixes.

**1. A popup that has already closed still claims the key.** `useOverlay`'s handler calls `stopPropagation()` and `preventDefault()` for *every* `Escape` and only then asks whether this overlay is the topmost one allowed to act on it — unlike `onInteractOutside`, which it gates on `isOpen` up front:

```js
let onKeyDown = (e) => {
  if (e.key === 'Escape' && !isKeyboardDismissDisabled && !e.nativeEvent.isComposing) {
    e.stopPropagation();
    e.preventDefault();
    onHide();            // <- only THIS checks the overlay stack
  }
};
```

Our popups stay mounted for their exit transition, so the closed popup consumes the key and then declines to use it. Instrumenting `Event.prototype.stopPropagation` in a real browser:

```
press #1  useOverlay owner=SelectOverlayWrapper isOpen=true isTop=false stack=1
```

`isTop=false` — the Modal is the only overlay on the stack, so `onHide()` no-ops.

Giving `useOverlay` an `isKeyboardDismissDisabled` of its own does **not** work: `DisplayTransition` replays the children it stored while the overlay was open, so every value that subtree captured is frozen at its open-state reading for exactly the window that needs guarding. `useOverlayEscapeGuard` works around it with `useEvent` — stable handler identity, latest closure in the body.

**2. Focus is left on a detaching option.** The popup's `FocusScope restoreFocus` only restores on unmount, at the *end* of the exit transition. Once the option detaches, focus drops to `<body>`, the event leaves the React tree, and no handler runs at all. `Select` now hands focus back to the trigger as soon as the list closes, guarded on focus still being inside the popup so clicking elsewhere is unaffected.

**3. The trigger's tooltip takes it.** `useTooltipTrigger` registers a *document-capture* keydown listener while open and `stopPropagation()`s `Escape`. Returning focus to a trigger (fix 2) opens its tooltip, which then eats the key.

A tooltip answers a user *arriving* at a trigger. Focus handed back by a component is not that arrival, but the DOM event is identical and the modality is still "keyboard", so React Aria shows it anyway. `focusProgrammatically` marks the difference — `focus()` dispatches synchronously, so a flag around the call is readable by every handler for it and gone immediately after, which is what makes this reliable where a timing window would not be — and `TooltipTrigger` declines to open for it.

Fixes 2 and 3 have to land together: 2 alone makes 3 fire reliably.

## Propagation, made uniform

`ListBox`, `Picker` and `FilterPicker` each stopped `Escape` leaving `useKeyboard` handlers that never used it (`useKeyboard` stops propagation for every key unless the handler opts back out; the `ListBox` comment — "let the overlay system handle closing" — shows only `preventDefault` was considered).

All three now continue propagation for **`Escape` alone**. Releasing every unhandled key would also let `Enter` and typeahead out of a list, which callers have always had contained — so the consistent rule here is the narrow one, not the broad one.

## Tests

`DialogTrigger.browser.test.tsx`, browser tier — the window does not exist in jsdom, which neither runs the exit transition nor keeps focus where the user left it.

- One Escape closes the Dialog right after a keyboard pick, a mouse pick, and a `Picker` pick, and after a nested popover is dismissed. **All four fail on `main`, all pass here.**
- The other half of the contract: an `Escape` a popup *can* act on still belongs to it alone — letting the key travel must not turn one press into two closes. Covered for `Select` and `Picker`.

Verified stable over 5 consecutive full-file runs (these cases failed ~1 run in 3 when only swallow 1 was fixed). Full suites green locally: 2725 jsdom, 188 browser, plus lint, `test:types`, `test:types:consumer`, `test:form-store`, `build` and the Form diagnostics ratchet.

## Note for the consumer side

This should remove the need for `dismissWithEscape()`'s re-send in cubedevinc/cubejs-enterprise#15167, but I have not run that suite — worth confirming against a real probation run before retiring it.


src/components/fields/ListBox/ListBox.tsx
```diff
@@ -859,10 +859,23 @@ export const ListBox = forwardRef(function ListBox<T extends object>(
...
+      //
+      // Deliberately `Escape` alone. Releasing every key this handler ignores
+      // would also let `Enter`, typeahead and the rest out, which callers have
+      // always had contained here; dismissal is the one key that is meaningless
+      // to hold on to.
+      //
+      // "Let the overlay system handle closing" is what the old comment here
+      // intended; only `preventDefault` was ever covered.
+      if (e.key === 'Escape') {
+        e.continuePropagation();
```
tenphi: [P2] Keep the propagation opt-out local to the Escape event

In the pinned `@react-aria/interactions@3.26.0`, `createEventHandler` stores `shouldStopPropagation` outside the returned event callback. Calling `continuePropagation()` therefore leaves propagation enabled for subsequent keys until this handler is recreated by a render. Reproduced in Chromium with a standalone ListBox containing Blue/Red and no selection: focus it, press Escape, then `z`; a parent `onKeyDown` receives both Escape and `z`. The same check with the base ListBox implementation contains `z`. This lets later typing invoke ancestor shortcuts while focus is still in the list, contrary to the intended Escape-only change. Make the propagation decision independently for each event.

src/components/fields/Picker/Picker.tsx
```diff
@@ -494,7 +494,20 @@ export const Picker = forwardRef(function Picker<T extends object>(
...
+      // `useKeyboard` stops propagation for every key it sees unless the
+      // handler opts back out, and this handler never acts on `Escape`. That
+      // key has to keep travelling: it is how whatever surrounds this — a
+      // Dialog, most often — gets dismissed (CUB-4839).
+      //
+      // Deliberately `Escape` alone. Releasing every key this handler ignores
+      // would also let `Enter`, typeahead and the rest out, which callers have
+      // always had contained here; dismissal is the one key that is meaningless
+      // to hold on to.
+      if (e.key === 'Escape') e.continuePropagation();
```
tenphi: [P2] Release Escape through ItemButton's outer keyboard wrapper

This `keyboardProps.onKeyDown` is passed to `ItemButton`, whose `useAction` → `useButton` → `useFocusable` adds another `useKeyboard` wrapper. `continuePropagation()` here releases only the inner wrapper; the outer one still stops Escape. Reproduced in Chromium with a Picker inside a modal Dialog and `triggerTooltip={false}`: select an option by keyboard, wait for the popup to unmount and focus to return to the trigger, then press Escape—the Dialog remains open. FilterPicker has the same problem. The new test presses Escape immediately while focus is still in the exiting popup, so it misses this path. Avoid the double wrapping or release propagation at the outer wrapper, and cover Escape after focus restoration.

tenphi (commented):
Reviewed `9f3162b`. Found one new keyboard-propagation regression and two remaining Escape failures in the scenarios this PR addresses; details and reproductions are inline.

Validation: all 15 existing `DialogTrigger.browser.test.tsx` tests passed, as did 114 targeted jsdom tests (1 skipped). Additional local Chromium cases reproduced the findings. The ListBox regression case passes with the base ListBox implementation and fails with this PR's implementation.

src/components/overlays/Tooltip/TooltipTrigger.tsx
```diff
@@ -160,6 +163,23 @@ export function TooltipTrigger(props: CubeTooltipTriggerProps) {
...
+  // component — a popup returning it to the trigger it came from — is not that
+  // arrival, but the DOM event is identical and the modality is still
+  // "keyboard", so React Aria would show the tooltip anyway. It then absorbs
+  // the user's next `Escape` from its own document-level listener, which is
+  // how a Dialog ended up needing two presses to close (CUB-4839).
+  //
+  // Dropping the focus handler rather than closing after the fact: React Aria
+  // decides to show inside it, so this never starts and nothing flashes.
+  const onFocus = useEvent((event: FocusEvent<HTMLElement>) => {
+    if (isProgrammaticFocus()) return;
```
tenphi: [P2] Cover FocusScope restoration in the tooltip suppression

Only Select calls `focusProgrammatically`; popover DialogTriggers (including Picker/FilterPicker) still restore focus through React Aria's `FocusScope`, so this guard sees `false` for those restores. Reproduced on this head in Chromium: put a popover DialogTrigger with `<Button tooltip="More options">` inside a modal Dialog, open the popover by keyboard, dismiss it with Escape, wait for it to unmount and focus to return, then press Escape again. The restored trigger opens its tooltip, whose document-capture listener consumes that second Escape, leaving the modal open. This is independent of the Picker keyboard-wrapper issue and reproduces with a plain Button. Include these restoration paths in the suppression and test after the exit animation completes.

src/components/fields/ListBox/ListBox.tsx
tenphi: Confirmed and fixed — this was a real regression from my change, and your repro is exactly right.

`createEventHandler` declares `shouldStopPropagation` in the factory closure and never resets it per event, so the opt-out survived until a render replaced the handler. Reproduced before fixing: focus a ListBox, `Escape`, then `z` → the parent saw both.

The decision is now made per event in `allowEscapeToPropagate`. One wrinkle worth flagging, because my first attempt at this was wrong in a way that was easy to miss: this handler is reached with **two different event shapes** — augmented by `useKeyboard`, and plain via ui-kit`&apos;`s `chain` — and the default is *opposite* between them. Re-asserting `stopPropagation()` unconditionally silently stopped the real event on the plain path. The helper branches on the shape.

Covered by `ListBox.browser.test.tsx`: Escape gets out, an ordinary key does not, and neither does a key pressed straight after Escape.

src/components/fields/Picker/Picker.tsx
tenphi: Confirmed and fixed, for both `Picker` and `FilterPicker`.

Took the "avoid the double wrapping" option: they now hand `ItemButton` a plain `onKeyDown` instead of `useKeyboard` output, so ours *is* the outer wrapper`&apos;`s handler and the opt-out applies to the event that actually travels.

You were also right that the existing test missed the path. Added `closes the Dialog on Escape once focus is back on the Picker trigger`, which waits for the popup to unmount and focus to return before pressing Escape — it fails without this change.

One correction to the diagnosis while I was in there: removing the double wrap alone did not fix it. With the inner wrapper gone the handler also gets invoked with a plain React event via ui-kit`&apos;`s `chain`, where `continuePropagation` does not exist and propagation is the default — so the per-event reset from the other thread was stopping Escape itself. Both threads needed the same shape-aware helper before either behaved.

src/components/overlays/Tooltip/TooltipTrigger.tsx
tenphi: Confirmed and fixed. You are right that the guard only ever saw `Select` — everything restoring through React Aria`&apos;`s `FocusScope` bypassed it, and your plain-Button repro reproduces on the old head.

A `FocusScope` restores from inside React`&apos;`s own commit with nothing to wrap, so there is no synchronous marker available for that path. `Overlay` now declares the restore window as it starts closing (`markFocusRestoreWindow`, exit duration + margin) and `TooltipTrigger` checks both that and the exact `focusProgrammatically` marker.

The time-based half is the weaker of the two and I would rather say so than hide it: it can only suppress a tooltip for a focus landing while an overlay is going away, so the worst case is a user who Tabs onto a trigger within ~550ms of closing an overlay not getting a tooltip until they move focus again. If you would prefer `Dialog` to own its restore explicitly instead — `restoreFocus={false}` plus a `focusProgrammatically` call — I am happy to switch, but that touches the CUB-3962 focus hand-off, so I did not do it unasked.

Covered by `closes the Dialog on Escape after a tooltip-bearing popover trigger is restored`, using `<Button tooltip="More options">` as in your repro.

tenphi: Re-reviewed `96e7d939`. The original reproduction now passes with normal motion, but this finding still reproduces with `prefers-reduced-motion: reduce`.

[P2] Establish tooltip suppression even when the exit phase is skipped

`Overlay.tsx:69` calls `markFocusRestoreWindow` only for the `exit` phase. With reduced motion, `DisplayTransition` sets its duration to zero: `ensureExitPendingFlow` sets `exit` and then synchronously sets `unmounted` in the same callback, so React batches away the `exit` render and its phase notification. The suppression window is never established before `FocusScope` restores focus.

I ran the new `tooltip-bearing popover trigger` test in a fresh Chromium context with `playwright({ contextOptions: { reducedMotion: 'reduce' } })`; it fails. An independent reproduction confirms that the inner popover unmounts, focus returns to its Button, `More options` opens as a tooltip, and the next Escape leaves the outer Dialog open. The same cases pass with normal motion. Please establish the suppression on a close/restore path that also runs without an exit animation, and cover reduced motion.

tenphi (commented):
Re-reviewed `96e7d939`. Verified and resolved the ListBox propagation and Picker/FilterPicker keyboard-wrapper findings. The tooltip finding is fixed with normal motion but remains open for reduced motion: `DisplayTransition` skips the `exit` notification, so `markFocusRestoreWindow` is never called and the restored tooltip consumes Escape. Reproduction and diagnosis are in the [existing tooltip thread](https://github.com/cube-js/cube-ui-kit/pull/1415#discussion_r4062081689).

Validation: 27 browser tests passed with normal motion, including the previous independent reproductions; 114 targeted jsdom tests passed (1 skipped). The PR's tooltip restoration test and an independent reproduction both fail in a fresh Chromium context with `reducedMotion: 'reduce'`. No other findings.

src/components/overlays/Tooltip/TooltipTrigger.tsx
tenphi: Confirmed and fixed. Good catch — this is exactly the failure mode a time-window carries, and I had hung it off the one signal that is not guaranteed.

Traced it to the same place you did: `ensureExitPendingFlow` sets `exit` and `ensureExitFlow` sets `unmounted` in the same callback once `dur` is `0`, so the `exit` render — and `onPhaseChange` with it — is batched away.

`Overlay` now opens the window on `isOpen` going false rather than on the exit phase. That signal always arrives, animation or not, and it is observed during render so it lands before the commit that unmounts the scope.

Reproduced your failure locally first by stubbing `matchMedia` for `prefers-reduced-motion` rather than adding a second browser project, so the case runs in the normal suite: `closes the Dialog on Escape after a restore, with reduced motion`. Verified it fails with the fix disabled and passes with it.

The tradeoff I flagged earlier is unchanged and still the weaker half of this: closing an overlay suppresses a focus-opened tooltip for ~550ms. The offer stands to switch `Dialog` to owning its restore explicitly if you would rather have an exact signal than a window — say the word and I will do it as a follow-up rather than widen this PR.

tenphi: Verified fixed on `9ffe8076`. The independent reproduction now passes in a fresh Chromium context with native `reducedMotion: 'reduce'`, as do the other six independent regression cases under that setting. The normal-motion browser suite, including the new reduced-motion test, also passes. Moving the marker to logical close covers the skipped-exit path. Resolving this finding.

tenphi (commented):
Re-reviewed `9ffe8076`: no new actionable findings. Verified the reduced-motion fix and resolved the last open finding; all findings from the earlier reviews are now addressed.

Validation: 28 browser tests passed under the standard configuration, 7 independent regression tests passed in a fresh Chromium context with native `reducedMotion: 'reduce'`, and 31 targeted overlay/transition/tooltip unit tests passed. CI's main test jobs were still running at the time of this review.

PR #1418, opened 2026-09-22
feat(Form): prepare modern dialogs, examples, and compiler coverage

tenphi:
Passing a modern controller to `DialogForm` now selects the modern form implementation, with inferred callbacks, automatic closing after successful saves, and cancellation when the dialog closes. Existing legacy callers keep their callback and reset contracts.

The ordinary Storybook form examples now use modern controllers. The form, dialog, and input documentation recommend modern forms for new work and explain how both implementations coexist; dedicated legacy component references remain available.

- Handle hidden dialogs, controlled closing, drafts, controller replacement, Strict Mode, and late async responses without resetting or closing a newly opened session. Popover actions stay open until saving succeeds; detached controllers reset using their own committed preserve policy.
- Let explicit Submit/Cancel actions close a `DialogTrigger` when passive dismissal is disabled.
- Add dialog coverage to the React Compiler suite, native Chromium interaction tests, legacy compatibility tests, and built-package type fixtures. Include the migrated stories in the source type-check gate.
- Include a minor changeset. This PR is scoped to UI Kit; Cloud adoption follows separately.

Validation: full unit suite (2,815 passed, 1 skipped), React Compiler suite (143 tests), isolated store suite (126 tests), targeted Chromium suite (26 tests), source and built-consumer type checks, package/Storybook/docs builds, lint, docs audit, compiler diagnostics ratchet, and Chromatic snapshot budget. Migrated stories were rendered in Chromium with their play functions. Bundle size remains within the existing limits. The existing CI matrix also verifies the modern form and compiled suites on React 18 and 19.

Expected visual changes: validation stories now show their error states, the DialogForm story opens its dialog before capture, and complex forms distinguish the range slider label. Chromatic baseline acceptance is left for review.


src/components/overlays/Dialog/ModernDialogForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+          autoComplete={autoComplete}
+        >
+          {typeof children === 'function' ? children(dismiss) : children}
+          {!noActions && (
+            <ButtonGroup>
+              <SubmitButton
+                qa={`${qa || ''}SubmitButton`}
+                theme={danger ? 'danger' : undefined}
+                label={t('dialogForm.submit', 'Submit')}
+                {...submitProps}
```
tenphi: [P2] Prevent Submit from dismissing a popover before saving

With `<DialogTrigger type="popover">` and an asynchronous `onSubmit`, clicking this default Submit button immediately closes the popover and aborts the save's signal. `Button` emits its default parent-popover dismissal on press, so the new closing effect calls `finishSession()` before the save resolves; values are also reset when `preserve` is false. I reproduced this in both jsdom and Chromium with a valid text field and a pending save. Opt the form/submit action out of automatic popover dismissal (for example, through `data-popover-keep`) so the successful-submission observer controls when it closes.

src/components/overlays/Dialog/ModernDialogForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+    if (!isClosing) sessionEnded.current = false;
+  }, [isClosing, form, finishSession]);
+
+  // Portals may relocate (remounting this component) and Strict Mode reconnects
+  // effects. A replacement root owns the same session; only a detached
+  // controller should reset after real removal.
+  useLayoutEffect(
+    () => () => {
+      queueMicrotask(() => {
+        if (!internals.getRootElement()) finishSession(form);
```
tenphi: [P2] Reset the detached controller when the form prop changes

Editing controller A and rerendering the same open `DialogForm` with controller B leaves A's edited values intact even with the default `preserve={false}`. This cleanup captures A, but `finishSession` is a `useEvent` callback and has already switched to the render for B by the time the microtask runs; its `target !== form` guard therefore skips A's reset. Reusing A later restores the discarded edits, unlike unmounting its dialog, which resets it. Keep cleanup/session ownership per controller so the detached controller is reset without touching the new controller.

tenphi (commented):
Found two reproducible issues in the new modern dialog lifecycle; details are inline.

Validation: the 24 existing modern/legacy dialog tests, 21 compiled dialog tests, and 126 isolated store tests pass. Additional regression cases reproduce both findings; the premature popover dismissal also reproduces in Chromium.

src/components/overlays/Dialog/ModernDialogForm.tsx
tenphi: Verified fixed in 4f0f2a3. The form-level `data-popover-keep` prevents premature dismissal for both built-in and custom submit actions. My original reproduction now passes, and the Chromium tests confirm that pending saves remain active until success while Cancel still aborts them.

src/components/overlays/Dialog/ModernDialogForm.tsx
tenphi: Verified fixed in 4f0f2a3. Cleanup now captures the detached controller's session and its committed preserve policy. My original controller-replacement reproduction passes, along with the new tests that swap controllers with opposite preserve settings and return to the old controller.

tenphi (commented):
Re-reviewed 4f0f2a3 against the previous review. Both reported issues are fixed, and I found no new actionable issues in the follow-up changes. Verified and resolved both original review threads.

Local validation passed: 28 modern/legacy dialog tests, 25 React Compiler dialog tests, 4 Chromium dialog tests, and both original review reproductions. The new coverage includes built-in/custom popover actions, failed and pending saves, explicit cancellation, and controller replacement with different preserve policies.

At posting time, the React 18/19 CI jobs are green; the full Tests & lint job, Browser tests, and Chromatic checks are still running or pending.

